PR #190, opened 2025-09-27
feat(rkl&libfuse-fs): add image unpack functionality for rkl and  adjust the log level in libfuse-fs

erasernoob:
This PR introduce three  mainly changes:
1. Run container from a local image directory
    - Implements the functionality for rkl to run containers directly from a local image directory following the OCI standard.(convert it to bundle dir)
    - Uses `libfuse-fs` to mount image layers instead of relying on the `MOUNT` syscall.
2. Replaces the `println!()` calls with `debug!()` macro for better log control in `libfuse-fs`.
3. Replaces `log` crate with `tracing` in `libfuse-fs`.


Ivanbeethoven (approved):
LGTM

PR #258, opened 2025-11-07
refactor: redesign the interfaces of BlockStore

mon3stera:
1. In original trait `BlockStore`, these methods need a `chunk_id` parameter to determined the block will be written. But `BlockStore` don't need to know what is chunk, which will let `BlockStore` more general and simpler. 
2. Let all methods of `BlockStore` receiver `&self` to enhance the degree of parallelism.


Ivanbeethoven (approved):
LGTM

PR #260, opened 2025-11-07
feat(slayerfs): use etcd watch and transactions  for consistency

yyjeqhc:
Use etcd watch to synchronize the file directory structure across multiple clients, and use etcd transactions to operate on the directory


Ivanbeethoven (approved):
LGTM

PR #264, opened 2025-11-10
FIX：the error occur on 087

briteny-pwn:


Ricky-Daxia (approved):
generic/087 tests `setattr` by calling `utime(2).

PR #379, opened 2026-01-23
Slayerfs enhancement: Implement buffer size limit to read path

mon3stera:
1. The current slayerfs don't have buffer size limit to read path, which will cause OOM in large read pattern. The PR implement a limit logic to it.
2. Add optional `jemalloc` feature, which allows us to use famous jemalloc allocator to enable advances features such as heap analysis.
3. Replace `std::sync::Mutex` with `parking_lot::Mutex` for better performance (really?)


genedna:
@codex review

PR #388, opened 2026-01-24
Slayerfs zerocopy on write-path: Implement best-effort zero-copy ability to write-path

mon3stera:
The PR introduces new implementation of page cache and object store backend. In current write-path, there is only one copy operation before sending bytes to backend, which can enhance the throughput by approximate 33%.


genedna:
@codex review

PR #407, opened 2026-01-30
Slayerfs profile: add profiling document and enhance write path

mon3stera:
1. Introduces `tracing-chrome` support and finish a document to describe how to use the supported profiling methods.
2. The result of benchmark shows that the block cache management is a critical bottleneck on the read path, consuming over half of the time. Given that the block cache is being refactored, I removed the related code in `ObjectBlockStore` and will wait for integration once it is ready. Now the read throughput is now twice as fast as before.  
3. In the previous VFS implementation, each write operation would call `extend_file_size` once, which was the critical bottleneck on write path. It has been postponed to when a slice is committing.
4. Introduces a new `TxnBuilder` for etcd. The `atomic_update` is quite good, but it cannot handle transactions requiring coordination across multiple keys. Although TxnBuilder is slightly more tedious to use, it is more general and better suited for subsequent metadata refactoring with `rkyv`.
5. A known defect is a slice upload task does not start unless it reaches the chunk end, which may cause OOM under a heavy write sequence. A better approach is starting the task as soon as a block is available and upload in blocks, which will significantly reduce memory pressure. (**Solved because of the new backpress mechanism**.)

Current performance metrics for reference:

- Single thread sequence write (4G): 1200MB/s.
- Multiple threads sequence write (16 threads/4G per file): 3100MB/s and 800MB memory usage.
- Single thread sequence read (4G): 1200MB/s.
- Multiple threads sequence read (16 threads/4G per file): 2100MB/s and 800MB memory usage. (which can enhance by 33% if there is a good block cache.)

### DIFF
--- a/project/slayerfs/benches/slayerfs_bench.rs
+++ b/project/slayerfs/benches/slayerfs_bench.rs
@@ -1,7 +1,7 @@
 use std::fs;
 use std::future::Future;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Once};
+use std::sync::{Arc, Once, OnceLock};
 use std::time::{Duration, Instant, SystemTime};
 
 use anyhow::{Context, Result, anyhow};
@@ -32,6 +32,7 @@ const MB: usize = 1024 * 1024;
 const KB: usize = 1024;
 
 static TRACING_INIT: Once = Once::new();
+static CHROME_GUARD: OnceLock<tracing_chrome::FlushGuard> = OnceLock::new();
 
 fn init_tracing(chrome_trace: Option<&Path>) {
     TRACING_INIT.call_once(|| {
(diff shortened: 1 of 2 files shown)


genedna:
@codex review

Ivanbeethoven:
> 4. Introduces a new `TxnBuilder` for etcd. The `atomic_update` is quite good, but it cannot handle transactions requiring coordination across multiple keys. Although TxnBuilder is slightly more tedious to use, it is more general and better suited for subsequent metadata refactoring with `rkyv`.为 etcd 引入了新的 `TxnBuilder`。`atomic_update` 相当不错，但无法处理需要跨多个密钥协调的交易。虽然 TxnBuilder 使用起来稍微繁琐一些，但它更通用，更适合后续用 `rkyv` 进行元数据重构。

@lei-jia-xing

PR #408, opened 2026-01-30
feat(slayerfs): impl async cache insert and add weigher

zineyu:
slayerfs的运行时配置将单独处理，因为它目前缺乏统一的配置方法

### DIFF
--- a/project/slayerfs/src/chuck/chunk_cache/cache.rs
+++ b/project/slayerfs/src/chuck/chunk_cache/cache.rs
@@ -119,6 +119,7 @@ impl Default for ChunksCacheConfig {
     }
 }
 
+#[allow(dead_code)]
 #[derive(Clone)]
 pub struct ChunksCache {
     /// Persistent disk storage backend with SHA256-based file naming
@@ -139,6 +140,7 @@ pub struct ChunksCache {
     config: ChunksCacheConfig,
 }
 
+#[allow(dead_code)]
 impl ChunksCache {
     /// Creates a new ChunksCache with default configuration
     #[allow(dead_code)]


genedna:
@codex review

genedna:
@claude

project/slayerfs/src/chuck/chunk_cache/cache.rs
```diff
@@ -0,0 +1,326 @@
...
+            .take()
+            .unwrap_or_else(|| cache_dir().unwrap());
+        debug!("Using cache directory: {:?}", cache_dir);
+        let disk_storage = DiskStorage::new(cache_dir).await?;
+
+        let hot_bytes = Arc::new(AtomicU64::new(0));
+        let hot_bytes_evict = hot_bytes.clone();
+        let hot_cache_builder = moka::future::Cache::builder()
+            .weigher(|_: &String, v: &Vec<u8>| v.len() as u32)
+            .max_capacity(config.hot_cache_max_bytes as u64)
```
mon3stera, line 171: `max_capacity` is not a hard limit in moka. In fact, according to moka's document, the evictions only occur in two situations:
1. There are at least 64 read/write records, such as `get/get_with` (read records) and `insert/invalidate` (write records).
2. At least 300ms have passed since the last maintenance.

It means that if there is a heavy read sequence, it is highly likely to get OOM. To be honest, I think using moka here is too heavy for us. The facts have proven that managing block cache may consume more resources than read/write themselves.

mon3stera:
The bcache is not as useful as we think. In contrast, maintaining the cache itself is resource-intensive. Especially in sequential read operations, once a block is read, it may not be read again. Therefore, we think that we need to change our policy:

1. The speed of true remote object storage is much slower than local storage. So it is acceptable to use disk instead of memory as cache. In default, we don't use memory as the primary cache, but rather as a "buffer for waiting to be written to disk". **The block to be cached first enter the pending queue with an upper limit of approximate 15 blocks, then are asynchronously flushed to disk**. In this simpler structure, we have not need to such a heavy memory cache like moka.
2. Some object storages provide range get interfaces. If reader requests a small block (`length < block_size`), **block cache can use range get interfaces to fetch the data directly and then initiate a prefetch request to obtain the full block**. 
3. In default, when performs a put operation, **we just need to cache blocks whose length is less than a full block instead of caching all blocks unconditionally**.
4. We need a configurable "writeback" policy. In default, the blocks may be cached and then are uploaded to object storage directly. In `writeback` policy, however, **the blocks are flushed to disk and asynchronously uploaded to object storage**. This sacrifices consistency in exchange for lower latency and higher throughput. 

Read Path:

```
  App Read
     │
     ▼
  FileReader
     │
     ▼
  background_fetch
     │
     ├─► 1) bcache.get(key)
     │        │
     │        ├─ hit in memory pending pages  ──► copy to user buf
     │        │
     │        └─ miss → try disk cache file ──► copy to user buf
     │
     └─► 2) cache miss:
              ├─ small range + seekable → Range GET
              │       └─ optional prefetch → bcache.insert(key)
              │
              └─ full block read (store.load)
                      ├─ cache_full_block=true (configuration) OR small block (length < block_size)
                      │       └─ bcache.insert(key)
                      └─ else no cache
              ▼
          copy to user buf

```

Write Path:

```
  App Write
     │
     ▼
  FileWriter
     │
     ▼
  SliceState
     │
     ├─► upload block
     │     │
     │     ├─ writeback=false (default)
     │     │      ├─ upload to object storage
     │     │      └─ if small block (length < block_size) OR cache_large_write (configuration)
     │     │             → bcache.insert(key)
     │     │
     │     └─ writeback=true
     │            ├─ stage() → write to disk cache dir (rawstaging)
     │            ├─ optionally link to cache (read cache index)
     │            └─ async upload later
     │
     ▼
```

block cache structure:

```
  bcache.insert(key, page)
     │
     ├─ pages[pending] (temporary buffer in memory)
     │      └─ flush() → write cache file to disk (main cache)
     │                    └─ update cache.keys index
     │
     └─ eviction applies to disk files (keys index)
```

zineyu:
## 1. Core Data Structures

### 1.1 Per-File Metadata `cache_inode`
- **Radix Tree `cache_tree`**: offset → `cache_entry` (fast lookup)
- **Doubly Linked LRU Lists**: `active_list`, `inactive_list` (cold/hot isolation for disk space reclamation)
- **Dirty Page List `dirty_list`**: Used only in Write-Back mode
- **Consistency Metadata**: `cached_version` (etag/version), `last_validate_time` (TTL mode)

### 1.2 Per-Cache Block `cache_entry`
- **Disk Path**: `<cache_root>/<file_id>/<offset>.blk` (fixed-size block, e.g., 64KB)
- **Offset, Block Size**, **Reference Count** (in-memory access count)
- **Status Flags**: `DIRTY`, `WRITEBACK`, `VALID` (data persisted to disk)
- **Validation Fields**: `version`/`etag`, `validate_ts`
- **Last Access Time** (used for LRU reclamation)

---

## 2. Cache Admission and Prefetching

### 2.1 Read Path
1. **Disk Data Hit**:
   - Read data from disk into memory buffer according to `cache_entry` disk path.
   - Update `last_access`, promote in LRU lists.
2. **Complete Miss**:
   - Synchronously initiate server RPC to read data.
   - Allocate disk block path, **write data to disk file synchronously** (atomic creation).
   - Create `cache_entry`, insert into radix tree, append to `inactive_list` tail.
   - Read data into memory buffer and return (optional: add to hot cache).

### 2.2 Adaptive Prefetching
- **Sequential Detection**: Exponential window growth (16KB → 1MB), fast backoff on random access.
- **Batched RPC**: Merge adjacent offset prefetch requests.
- **Asynchronous Persistence**: Prefetched data is written to disk in the background; `cache_entry` is created and added to `inactive_list` upon completion.
- **Prefetched blocks are not placed into the hot cache** (unless accessed) to avoid memory pollution.

---

## 3. Cache Replacement Policy (Disk Space Reclamation)

### 3.1 Space Quota (Pure In-Memory Control)
- **Hard Limit**: `max_disk_cache_size` (e.g., 50GB), **full metadata in memory tracks used space**.
- **Soft Watermark**: 80% → background reclamation thread triggers.

### 3.2 Reclamation Algorithm (Double-list LRU + Second Chance)
- **Cold/Hot Isolation**:
  - New blocks are added to the tail of `inactive`.
  - Accessed once → promoted to `active`.
  - `active` blocks not used for a long time are demoted back to `inactive`.
- **Reclamation Scan**:
  - Scan only the tail of `inactive`.
  - **Skip**: reference count > 0 (currently in memory use) or `DIRTY` state.
  - **Dirty Blocks**: Trigger asynchronous write-back; upon completion clear `DIRTY` and reinsert at the tail of the list.
  - **Clean Blocks**: Delete disk file, remove `cache_entry`, decrease total usage counter.

### 3.3 Direct Reclamation
- When allocating a new block and the hard limit is reached, the current thread blocks and participates in reclamation (synchronously evicts blocks from the tail of `inactive`).

---

## 4. Write Cache Policy

### 4.1 Default Mode: Write-Through
- **Flow**: Write request is sent synchronously to the server → upon success, write to local disk file → create/update `cache_entry`.
- **No dirty pages**, natural "last writer wins" across multiple clients.
- **Optimization**: Small write coalescing (delayed merging), but still requires server acknowledgment.

### 4.2 Optional Mode: Write-Back
- **Flow**: Data is written directly to the disk file, marked `DIRTY` → return success immediately.
- **Persistence Guarantee**: Disk is non-volatile; dirty pages **still exist on disk** after client crash, but **discarded on restart** (metadata lost, cannot be recovered).
- **Flush Policy**: Background thread periodically writes back to server; clears `DIRTY` upon success.
- **Conflict Handling**: Requires business layer to ensure single writer, or application-level eventual consistency.

---

## 5. Concurrency and Lock Optimization

### 5.1 Metadata Access (In-Memory)
- **RCU-protected Read Path**: Radix tree lookup performed within `rcu_read_lock()`, completely lock-free.
- **Write Operations** (insert/delete/update): Use `spinlock` + `call_rcu` to defer node freeing.
- **Segmented LRU Locks**: Partitioned into multiple buckets based on file ID hash to reduce global contention.

### 5.2 Disk I/O Concurrency
- **Asynchronous I/O** (libaio/io_uring) for concurrent disk reads/writes.
- **Request Merging**: Adjacent offset read requests are merged into a single disk read (≤1MB).
- **Write Buffering**: Dirty page write-back can be submitted in batches.

### 5.3 Hot Data Page Concurrency
- Page-level reference counting, RCU-protected lookup, independent LRU lock segmentation.

---

## 6. Memory Management and Quota Control

### 6.1 Memory Metadata Overhead
- Each `cache_entry` occupies approximately **64 bytes**, supporting millions of cache blocks (e.g., ~64GB cache).

### 6.2 Disk Space Quota (In-Memory Accounting)
- **Global Counter**: Total disk space currently occupied in bytes.
- **Allocation**: Check against hard limit → trigger direct reclamation if exceeded.
- **Deletion**: Atomically decrement the counter.

### 6.3 Background Reclamation Thread
- Periodically checks if disk usage exceeds the soft watermark → performs batch reclamation from the tail of the `inactive_list`.

---

@mon3stera

PR #409, opened 2026-02-01
Perf/meta optimize

lei-jia-xing:
BRIEF CHANGES:
- Add rkyv serialization foundation with feature flag and backward-compatible JSON fallback
- Migrate SliceDesc, Etcd store structs, Permission/ACL, FileLock types to rkyv
- Implement atomic Redis Lua scripts for metadata ops (CREATE_ENTRY, RMDIR, RENAME, RENAME_EXCHANGE, LINK, UNLINK)
- Add configurable etcd watch with fine-grained prefixes
- Enable KVM integration tests with TCG fallback support

link #398

### DIFF
--- a/project/slayerfs/src/fuse/mod.rs
+++ b/project/slayerfs/src/fuse/mod.rs
@@ -206,7 +206,7 @@ where
         })
     }
 
-    // Open file: stateless IO, always return fh=0
+    // Open file: allocate a handle for read/write operations.
     async fn open(&self, _req: Request, ino: u64, flags: u32) -> FuseResult<ReplyOpen> {
         // Verify the inode exists and is a file
         let Some(attr) = self.stat_ino(ino as i64).await else {
@@ -296,21 +296,25 @@ where
         })
     }
 
-    // Write file: inode-based write to avoid path resolution
     async fn write(
         &self,
         _req: Request,
         ino: u64,
-        _fh: u64,
+        fh: u64,
         offset: u64,
         data: &[u8],
         _write_flags: u32,
         _flags: u32,
     ) -> FuseResult<ReplyWrite> {
-        let n = self
-            .write_ino(ino as i64, offset, data)
-            .await
-            .map_err(Into::<Errno>::into)? as u32;
+        let n = if fh != 0 {
+            self.write(fh, offset, data)
+                .await
+                .map_err(Into::<Errno>::into)? as u32
+        } else {
+            self.write_ino(ino as i64, offset, data)
+                .await
+                .map_err(Into::<Errno>::into)? as u32
+        };
         Ok(ReplyWrite { written: n })
     }
 
@@ -706,7 +710,7 @@ where
         parent: u64,
         name: &OsStr,
         mode: u32,
-        _flags: u32,
+        flags: u32,
     ) -> FuseResult<ReplyCreated> {
         let name = name.to_string_lossy();
         // Validate parent
@@ -731,11 +735,20 @@ where
             return Err(libc::ENOENT.into());
         };
         let attr = vfs_to_fuse_attr(&vattr, &req);
+
+        let accmode = flags & (libc::O_ACCMODE as u32);
+        let read = accmode != (libc::O_WRONLY as u32);
+        let write = accmode != (libc::O_RDONLY as u32);
+        let fh = self
+            .open(ino, vattr.clone(), read, write)
+            .await
+            .map_err(Into::<Errno>::into)?;
+
         Ok(ReplyCreated {
             ttl: Duration::from_secs(1),
             attr,
             generation: 0,
-            fh: 0,
+            fh,
             flags: 0,
         })
     }


project/slayerfs/src/vfs/fs.rs
```diff
@@ -1958,6 +1958,12 @@ where
             .await
             .map_err(VfsError::from)?;
 
+        let target_size = offset + written as u64;
+        self.core
+            .meta_layer
+            .extend_file_size(ino, target_size)
```
mon3stera: This is redundant as the writer will update the file size by itself, so `VFS` has no need to manage it.

project/slayerfs/Cargo.toml
```diff
@@ -52,11 +52,13 @@ if-addrs = { workspace = true }
 uuid = { workspace = true, features = ["v7"] }
 parking_lot = { workspace = true }
 tikv-jemallocator = { workspace = true, optional = true }
+rkyv = { version = "0.8", optional = true, features = ["uuid-1"] }
```
mon3stera: Maybe we need to define dependencies in `Cargo.toml` in the `project/`.

project/slayerfs/src/meta/mod.rs
```diff
@@ -21,6 +21,7 @@ pub mod file_lock;
 pub mod layer;
 pub(crate) mod migrations;
 pub mod permission;
+pub mod serialization;
```
mon3stera: Should be pub(crate).

project/slayerfs/src/vfs/fs.rs
lei-jia-xing: ok, I will fix that, vfs should call only once to keep atomic operation lower to meta. but remove that cause in kvm test error, I wiil figure it out

lei-jia-xing: It may due to the spawn ,the vfs write is over, but meta update is not over

lei-jia-xing: @mon3stera For user , we hope when we write a file , and we can query correct file meta later at any time. instead of querying old meta data in a short period

lei-jia-xing: I think we should keep this diff for the moment, it keep the accuacy

lei-jia-xing (commented):
For user , we hope when we write a file , and we can query correct file meta later at any time. instead of querying old meta data in a short period. So it's a problem that need to be soloved. keep vfs extend_file_size make sense. beacause wirter uses spawn, can't await. We just need to ensure the meta operation is idempotent. For the moment, I keep it.
**Future optimization:**
Consider Write-Ahead Log (WAL) to achieve both crash-safety and low latency:
- Write metadata intents to local durable log (fast)
- Background replayer commits to database (batched)
- Maintains crash-safety without blocking write path

mon3stera:
For the same client on the local machine, the writer updates the cached file length in memory immediately after writing, so other local operations can see the latest file state at any time. Across different clients, however, we only guarantee **close-to-open** consistency: the latest metadata is guaranteed to be visible **only** after **the file is closed and reopened**. Therefore, asynchronously committing file-length updates is fully expected. Please note that we don't provide strong consistency across clients.

lei-jia-xing:
@mon3stera  So, what do we need to do ? Do you mean we should change our write behaviour ? Now write operatioin in meta is to write slices & extend_file_size. Do you mean we should remove all the extend_file_size in meta? Instead do in flush  to do lazy load ？I think it's reasonable

mon3stera:
The fact is that `extend_file_size` is just a historical leftover. Now the writer updates the file length when submitting a new slice by calling `MetaLayer::write`. `MetaLayer::write` actually performs `MetaLayer::append_slice + extend_file_size` atomically, so you can remove these redundant methods (`extend_file_size + append_slice`).

lei-jia-xing:
But It result in kvm test error..
    exec_check(
        vm,
        &format!("sh -lc 'echo hello > {}/a'", SLAYERFS_MOUNTPOINT),
    )
    .await?;
    exec_check(
        vm,
        &format!("sh -lc 'cat {}/a | grep -q hello'", SLAYERFS_MOUNTPOINT),
    )

just a simple test can trigger. It is no matter?

genedna:
@claude

mon3stera:
> But It result in kvm test error.. exec_check( vm, &format!("sh -lc 'echo hello > {}/a'", SLAYERFS_MOUNTPOINT), ) .await?; exec_check( vm, &format!("sh -lc 'cat {}/a | grep -q hello'", SLAYERFS_MOUNTPOINT), )
> 
> just a simple test can trigger. It is no matter?

This is a bug in fuse adapter instead of VFS. The `open` in fuse adapter don't truly open a handle, which cause the automatically flush operations when closing a handle don't happen. Therefore the read may not see the newest data. Please patch the fuse adapter:

```rust
  // Open file: allocate a handle for read/write operations.
  async fn open(&self, _req: Request, ino: u64, flags: u32) -> FuseResult<ReplyOpen> {
      // Verify the inode exists and is a file
      let Some(attr) = self.stat_ino(ino as i64).await else {
          return Err(libc::ENOENT.into());
      };
      if matches!(attr.kind, VfsFileType::Dir) {
          return Err(libc::EISDIR.into());
      }

      let accmode = flags & (libc::O_ACCMODE as u32);
      let read = accmode != (libc::O_WRONLY as u32);
      let write = accmode != (libc::O_RDONLY as u32);
      let fh = self
          .open(ino as i64, attr.clone(), read, write)
          .await
          .map_err(Into::<Errno>::into)?;

      Ok(ReplyOpen { fh, flags })
  }

  async fn write(
      &self,
      _req: Request,
      ino: u64,
      fh: u64,
      offset: u64,
      data: &[u8],
      _write_flags: u32,
      _flags: u32,
  ) -> FuseResult<ReplyWrite> {
      let n = if fh != 0 {
          self.write(fh, offset, data)
              .await
              .map_err(Into::<Errno>::into)? as u32
      } else {
          self.write_ino(ino as i64, offset, data)
              .await
              .map_err(Into::<Errno>::into)? as u32
      };
      Ok(ReplyWrite { written: n })
  }

```

lei-jia-xing:
Got it.But it do not works :( . 
I will work on it. I actually sure it's fuse issue

lei-jia-xing:
https://github.com/libfuse/libfuse/blob/master/include/fuse_lowlevel.h#L954-L985
I notice that our handle do not set the correct fd .create change to create & open

lei-jia-xing:
> > But It result in kvm test error.. exec_check( vm, &format!("sh -lc 'echo hello > {}/a'", SLAYERFS_MOUNTPOINT), ) .await?; exec_check( vm, &format!("sh -lc 'cat {}/a | grep -q hello'", SLAYERFS_MOUNTPOINT), )
> > just a simple test can trigger. It is no matter?
> 
> This is a bug in fuse adapter instead of VFS. The `open` in fuse adapter don't truly open a handle, which cause the automatically flush operations when closing a handle don't happen. Therefore the read may not see the newest data. Please patch the fuse adapter:
> 
> ```rust
>   // Open file: allocate a handle for read/write operations.
>   async fn open(&self, _req: Request, ino: u64, flags: u32) -> FuseResult<ReplyOpen> {
>       // Verify the inode exists and is a file
>       let Some(attr) = self.stat_ino(ino as i64).await else {
>           return Err(libc::ENOENT.into());
>       };
>       if matches!(attr.kind, VfsFileType::Dir) {
>           return Err(libc::EISDIR.into());
>       }
> 
>       let accmode = flags & (libc::O_ACCMODE as u32);
>       let read = accmode != (libc::O_WRONLY as u32);
>       let write = accmode != (libc::O_RDONLY as u32);
>       let fh = self
>           .open(ino as i64, attr.clone(), read, write)
>           .await
>           .map_err(Into::<Errno>::into)?;
> 
>       Ok(ReplyOpen { fh, flags })
>   }
> 
>   async fn write(
>       &self,
>       _req: Request,
>       ino: u64,
>       fh: u64,
>       offset: u64,
>       data: &[u8],
>       _write_flags: u32,
>       _flags: u32,
>   ) -> FuseResult<ReplyWrite> {
>       let n = if fh != 0 {
>           self.write(fh, offset, data)
>               .await
>               .map_err(Into::<Errno>::into)? as u32
>       } else {
>           self.write_ino(ino as i64, offset, data)
>               .await
>               .map_err(Into::<Errno>::into)? as u32
>       };
>       Ok(ReplyWrite { written: n })
>   }
> ```

By the way, I don't like write_ino fallback. @mon3stera

mon3stera:
LGTM

PR #452, opened 2026-02-12
feature(slayerfs) : add slice compact and gc for slayerfs

Tyooughtul:
feature : 为slayerfs增加了 slice compact机制和gc机制 #355 
- 参考juiceFS的方式按照slice数量进行清理，实现了合并/替换、删除
- 增加碎片率来避免无效的compact
- 首先标记为delayed，然后在gc的时候一并清理
- 定时gc

### DIFF
--- a/project/slayerfs/src/lib.rs
+++ b/project/slayerfs/src/lib.rs
@@ -28,7 +28,7 @@ pub use crate::cadapter::s3::{S3Backend, S3Config};
 pub use crate::chuck::chunk::ChunkLayout;
 pub use crate::chuck::store::{BlockKey, BlockStore, InMemoryBlockStore, ObjectBlockStore};
 pub use crate::meta::client::MetaClient;
-pub use crate::meta::config::{CacheConfig, ClientOptions, Config, DatabaseConfig, DatabaseType};
+pub use crate::meta::config::{CacheConfig, ClientOptions, CompactConfig, Config, DatabaseConfig, DatabaseType};
 pub use crate::meta::factory::MetaStoreFactory;
 pub use crate::meta::file_lock::{FileLockInfo, FileLockQuery, FileLockRange, FileLockType};
 pub use crate::meta::store::{


genedna:
@codex review

project/slayerfs/src/meta/stores/database_store.rs
```diff
@@ -1013,6 +1017,311 @@ impl DatabaseMetaStore {
             }
         }
     }
+
+    /// 删除指定chunk的所有切片（内部方法，接受事务参数）
```
mon3stera: All comments should be written in English.

project/slayerfs/src/meta/stores/etcd_store.rs
mon3stera, line 1: Is this PR complete? The methods related to compaction aren't implemented for Etcd and Redis.

project/slayerfs/src/meta/stores/database_store.rs
```diff
@@ -2836,12 +3146,250 @@ impl MetaStore for DatabaseMetaStore {
...
+    /// example:
+    /// - slice A: slice_id=1, offset=0, length=100 (old)
+    /// - slice B: slice_id=2, offset=50, length=100 (newer, covers 50-150)
+    /// - slice A's [50-100] is covered by B, but [0-50] is still needed
+    /// - result: keep both A and B (no fully covered slices)
+    ///
+    /// - slice A: slice_id=1, offset=0, length=100 (old)
+    /// - slice B: slice_id=2, offset=0, length=150 (newer, fully covers A)
+    /// - result: remove A, keep B
+    async fn merge_slices(&self, slices: &[SliceDesc]) -> Result<Vec<SliceDesc>, MetaError> {
```
mon3stera, line 3226: The behave of `merge_slices` is not satisfactory. We shouldn't just remove the slices which are completely covered by others. Instead, we need to generate a minimal single slice from a large amount of slices.  This is much complexer than current implementation. Please note that we need to write the data of the new slice into object storage. Therefore, metadata stores cannot do that just by themselves.

mon3stera (changes requested):

project/slayerfs/src/meta/stores/database_store.rs
```diff
@@ -2836,12 +3146,250 @@ impl MetaStore for DatabaseMetaStore {
...
+        let mut result_slices: Vec<SliceDesc> = Vec::new();
+
+        for slice in slices_sorted {
+            let slice_start = slice.offset;
+            let slice_end = slice.offset + slice.length;
+
+            // calculate uncovered portions of this slice
+            let mut remaining = vec![(slice_start, slice_end)];
+
+            for &(covered_start, covered_end) in &covered_ranges {
```
mon3stera, line 3250: We can do this with `utils::Intervals`

project/slayerfs/src/chuck/compactor.rs
```diff
@@ -0,0 +1,1136 @@
...
+    ) -> Result<(), CompactorError> {
+        self.meta_store
+            .replace_slices_for_compact(chunk_id, new_slices, delayed_data)
+            .await
+            .map_err(CompactorError::MetaError)?;
+
+        debug!(
+            chunk_id = chunk_id,
+            new_slice_count = new_slices.len(),
+            delayed_count = delayed_data.len() / 12,
```
genedna, line 382: **Bug**: The debug log below (line 504) computes `delayed_count = delayed_data.len() / 12`, but the actual format is 20 bytes per slice (8 + 8 + 4). This should be `/ 20`.

Also, `prepare_delayed_data` silently truncates `slice.length` to `u32`. If a slice is larger than 4GB, the GC will only delete a subset of blocks. Consider using `u64` for the size field, or at minimum add a `warn!` when truncation occurs.

project/slayerfs/src/chuck/compactor.rs
```diff
@@ -0,0 +1,1136 @@
...
+
+        // Calculate merged slice layout using Intervals
+        let merged_slices = self.calculate_merged_slices(&slices)?;
+
+        // Calculate fragmentation ratio: (total_size - merged_size) / total_size
+        let total_size: u64 = slices.iter().map(|s| s.length).sum();
+        let merged_size: u64 = merged_slices.iter().map(|s| s.length).sum();
+        let fragment_ratio = if total_size > 0 {
+            (total_size - merged_size) as f64 / total_size as f64
+        } else {
```
genedna, line 93: The `_compactor` parameter is unused — the function delegates entirely to `meta_store.run_compact_by_threshold()`. This makes the `Compactor` struct effectively unused in the background worker flow. Either:
1. Remove the parameter and simplify, or
2. Actually use `Compactor::compact_chunk` here (which does the real block-level data merge)

Currently `meta_store.run_compact_by_threshold()` only removes fully-covered slice metadata but does NOT merge the actual block data. The `Compactor` was designed to do the data merge but is never called.

project/slayerfs/src/chuck/compactor.rs
```diff
@@ -0,0 +1,1136 @@
...
+            debug!(
+                chunk_id = chunk_id,
+                "No slices are fully covered, skipping compaction"
+            );
+            return Ok(None);
+        }
+
+        // Read data from old slices and merge
+        let chunk_size = self.layout.chunk_size;
+        let mut merged_data = vec![0u8; chunk_size as usize];
```
genedna, line 139: **Performance**: `let mut merged_data = vec![0u8; chunk_size as usize]` — with default `chunk_size` of 64MB, this allocates 64MB per compaction call. If multiple chunks are being compacted, memory usage will spike.

Consider:
- Processing block-by-block instead of loading the entire chunk into memory
- Or at least capping the allocation to the actual data range (`max(slice.offset + slice.length)` instead of full `chunk_size`)

project/rfuse3/src/path/inode_path_bridge.rs
```diff
@@ -756,12 +751,20 @@ where
             let inode = if entry.name == OsStr::new(".") {
                 parent
             } else if entry.name == OsStr::new("..") {
-                self.inode_name_manager
-                    .get_parent_inode(parent)
-                    .unwrap_or(ROOT_INODE)
+                inode_name_manager
+                    .inode_to_names
+                    .get(&parent)
+                    .unwrap()
```
genedna, line 757: **Panic risk**: `.unwrap().iter().next().unwrap()` will panic if:
1. `inode_to_names` doesn't contain the `parent` key (e.g., if it was evicted via `forget`)
2. The names set is empty

The old code used `get_parent_inode()` which returned `Option` and fell back to `ROOT_INODE`. This should use safe access:
```rust
inode_name_manager
    .inode_to_names
    .get(&parent)
    .and_then(|names| names.iter().next().map(|n| n.parent))
    .unwrap_or(ROOT_INODE)
```
Same issue in `readdirplus` (~line 1068).

project/rfuse3/src/path/inode_path_bridge.rs
```diff
@@ -129,60 +96,41 @@ impl InodeNameManager {
...
-        actual_inode
-    }
-
-    /// Get parent inode for a given inode
-    fn get_parent_inode(&self, inode: Inode) -> Option<Inode> {
-        self.inode_to_names
-            .get(&inode)
-            .and_then(|names| names.iter().next().map(|n| n.parent))
+    fn get_name_inode(&self, name: &Name) -> Option<Inode> {
+        self.name_to_inode.get(name).copied()
```
genedna, line 100: **Performance concern**: Replacing `DashMap` with `RwLock<HashMap>` means every filesystem operation (lookup, getattr, read, write, etc.) now contends on a single RwLock. `DashMap` provided sharded concurrent access which is much better for FUSE workloads with many parallel operations.

If the dependency on `dashmap` was intentionally removed, consider:
- `scc::HashMap` (lock-free concurrent hashmap)
- Multiple `RwLock<HashMap>` shards keyed by `inode % N`
- At minimum, document this as a known regression

project/rfuse3/src/path/inode_path_bridge.rs
```diff
@@ -129,60 +96,41 @@ impl InodeNameManager {
...
     }
 }
 
 pub struct InodePathBridge<FS> {
     path_filesystem: FS,
-    /// Concurrent inode-name manager using DashMap internally
-    inode_name_manager: InodeNameManager,
+    inode_name_manager: RwLock<InodeNameManager>,
 }
```
genedna, line 107: **Dead code**: A `Slab` is created here, a dummy `()` is inserted to skip key 0, but the slab is never stored or used. This should be removed.

project/slayerfs/src/meta/stores/database_store.rs
```diff
@@ -1013,6 +1019,364 @@ impl DatabaseMetaStore {
...
+    async fn cleanup_delayed_slices(
+        &self,
+        chunk_id: u64,
+        delayed: &[u8],
+        txn: &DatabaseTransaction,
+    ) -> Result<(), MetaError> {
+        if delayed.is_empty() {
+            return Ok(());
+        }
+
```
genedna, line 1067: **Performance**: `run_compact_by_threshold` queries ALL distinct chunk_ids, then calls `should_compact_chunk` for each (which itself does 2+ queries). For a filesystem with thousands of chunks, this is O(N) database round-trips.

Replace with a single aggregate query:
```sql
SELECT chunk_id, COUNT(*) as cnt FROM slice_meta
GROUP BY chunk_id
HAVING cnt >= :min_slice_count
ORDER BY cnt DESC
LIMIT :max_chunks_per_run
```
This reduces the initial scan to 1 query and only processes chunks that actually need evaluation.

project/slayerfs/src/meta/stores/redis_store.rs
```diff
@@ -2073,6 +2073,195 @@ impl MetaStore for RedisMetaStore {
...
+                    end
+                end
+                if not is_old then
+                    table.insert(keep_list, item)
+                end
+            end
+            
+            redis.call('DEL', key)
+            
+            for _, item in ipairs(keep_list) do
```
genedna, line 2118: **Race condition**: The SCAN + GET + DEL sequence is not atomic. Between GET and DEL, another process could:
1. Read the same delayed slice
2. Delete the block data
3. Then this process also tries to delete

Consider using a Redis sorted set with timestamp as score:
```
ZADD delayed_slices <timestamp> <slice_data>
ZRANGEBYSCORE delayed_slices -inf <cutoff_time> LIMIT 0 <batch_size>
ZREMRANGEBYSCORE delayed_slices -inf <cutoff_time>
```
Or wrap in a Lua script for atomicity.

project/slayerfs/src/meta/stores/etcd_store.rs
```diff
@@ -3424,6 +3424,117 @@ impl MetaStore for EtcdMetaStore {
...
+            || Ok((new_slices.to_vec(), ())),
+            10,
+            &None,
+        )
+        .await?;
+
+        if !old_slices_to_delay.is_empty() {
+            let now = Utc::now().timestamp();
+
+            // Format: slice_id (u64) + offset (u64) + size (u32) = 20 bytes per slice
```
genedna, line 3454: **Silent error**: When `serde_json::from_str` fails on a delayed slice value (line 3454-3456), the code silently `continue`s. This could hide data corruption or format migration issues. Add a `warn!` log with the key and error.

project/rfuse3/src/raw/session/handlers.rs
mon3stera, line 1: It seems that you failed to properly resolve conflicts, causing rfuse to revert to an older version. Please check twice.

project/slayerfs/src/chuck/compactor.rs
```diff
@@ -0,0 +1,1136 @@
...
+        Ok(Some(new_slice_id))
+    }
+
+    /// Calculate merged slices by removing fully covered slices.
+    ///
+    /// This only removes slices that are **completely** covered by newer slices.
+    /// Partially covered slices are kept intact because changing a slice's
+    /// offset/length would break block data addressing — block data is stored
+    /// at `(slice_id, block_index)` relative to the slice's original offset.
+    fn calculate_merged_slices(
```
mon3stera, line 186: Same as the previous question, we should really merge overlapped slices into a single slice and skip the slices that aren't deserved to merge while just removing fully covered slices.

project/slayerfs/src/chuck/compaction_worker.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        (compaction_handle, gc_handle)
+    }
+}
+
+/// run a single compaction cycle.
+#[allow(dead_code)]
+async fn run_compaction_cycle<M, B>(
+    meta_store: &Arc<M>,
+    _compactor: &Arc<Compactor<M, B>>,
```
mon3stera, line 90: The `Compactor` is unused. Why does it matter?

project/slayerfs/src/meta/stores/database_store.rs
```diff
@@ -2836,115 +3201,498 @@ impl MetaStore for DatabaseMetaStore {
...
+    /// - slice A: slice_id=1, offset=0, length=100 (old)
+    /// - slice B: slice_id=2, offset=50, length=100 (newer)
+    /// - A's [50-100] is covered by B, but [0-50] is still needed
+    /// - since we can't split A without rewriting, we keep all of A
+    /// - result: [Slice(1, 0, 100), Slice(2, 50, 100)]
+    ///
+    /// - slice A: slice_id=1, offset=0, length=100 (old)
+    /// - slice B: slice_id=2, offset=0, length=150 (newer, fully covers A)
+    /// - result: [Slice(2, 0, 150)] (A is completely removed)
+    async fn merge_slices(&self, slices: &[SliceDesc]) -> Result<Vec<SliceDesc>, MetaError> {
```
mon3stera, line 3226: duplicate?

mon3stera:
I'm concerned that parts of this PR may have been accepted from AI. This is possible and encouraged, provided that you thoroughly review the code given by AI. Please review the implementation end-to-end, ensure you understand each part, and refactor the duplicated/messy sections before we merge. 

In addition, I've noticed that we don't have a consistent approach to compaction path. Considering that reviewing a big PR is hard and the slice compaction is a crucial mechanism for SlayerFS, could you define and implement the overall compaction-deletion data flow and implement it on a metadata backend first? After we ensure that the data flow is correct, you can submit the rest parts. I think this should be a better way.

project/rfuse3/src/raw/session/handlers.rs
Tyooughtul: > It seems that you failed to properly resolve conflicts, causing rfuse to revert to an older version. Please check twice.


I was debugging macOS compilation issues locally, and accidentally included rfuse3 changes in this commit. Another mentor confirmed we don't need macOS support currently soI revert all rfuse3 changes.

project/slayerfs/src/chuck/compaction_worker.rs
Tyooughtul: > The `Compactor` is unused. Why does it matter?

These are placeholders for block-level compaction. Currently, run_compaction_cycle only calls meta_store.run_compact_by_threshold(). The plan is to replace this with compactor.compact_chunk(), which does full block read-merge-write, and use config.max_chunks_per_run to control throughput at this layer. I'll add TODO comments to make the intent explicit.

project/slayerfs/src/chuck/compactor.rs
Tyooughtul: > Same as the previous question, we should really merge overlapped slices into a single slice and skip the slices that are deserved to merge while just removing fully covered slices.

I think bug here is fragmentation ratio calculation is flawed. Since both calculate_merged_slices() and merge_slices() only identify fully-covered slices. I'll fix it to use proper interval merging for accurate 
fragmentation measurement.

project/slayerfs/src/meta/stores/database_store.rs
Tyooughtul: > duplicate?

These two functions serve different layers, database_store::merge_slices is called by compact_chunk_with_delay, which is metadata-only path, while compactor::calculate_merged_slices is used by the block-level compaction path. I will reorganize the code later.

project/rfuse3/src/raw/session/handlers.rs
mon3stera: You need to open a new single PR to revert that.

project/slayerfs/src/chuck/compactor.rs
mon3stera: The current PR is a little messy. Could you arrange your code and open a new clear PR for it?

Tyooughtul:
> I'm concerned that parts of this PR may have been accepted from AI. This is possible and encouraged, provided that you thoroughly review the code given by AI. Please review the implementation end-to-end, ensure you understand each part, and refactor the duplicated/messy sections before we merge.
> 
> In addition, I've noticed that we don't have a consistent approach to compaction path. Considering that reviewing a big PR is hard and the slice compaction is a crucial mechanism for SlayerFS, could you define and implement the overall compaction-deletion data flow and implement it on a metadata backend first? After we ensure that the data flow is correct, you can submit the rest parts. I think this should be a better way.

Apologies for the rough state of this PR, I've been without a Linux environment for e2e testing  these days and was using the draft mainly to get CI feedback, which I should have flagged upfront. I've since set up a local CI pipeline so this won't happen again. Will clean everything up before re-requesting review.

PR #455, opened 2026-02-13
Slayerfs singleflight

Ivanbeethoven:


genedna:
@codex review

project/slayerfs/src/chuck/singleflight.rs
mon3stera, line 1: Can you demonstrate that the usage of `SingleFilight` follows Read-after-Wrtie consistency?

mon3stera (changes requested):

mon3stera:
The `SingleFlight` provides a method to merge a result reusage for read requests. However, that's precisely the problem. The write operation has no coordination with read operation. Consider a specific time sequence: 
1. The first read operation has be submitted and it is waiting for the data.
2. A write operation was submitted and successfully executed.
3. At this moment, the first read will get the stale data, which is expected because it submited the request before the successful write. However, if a new read operation happens before the first read completes, it will reuse the result of the first read, but it is incorrect because it should see the newest data. This violates `Read-after-Write Consistency` even in single client.

Reusing the results of concurrent read operations isn't a good idea. Because the precondition that we can do this is that we must ensure that there isn't any write operation between two different but concurrent read operations. It is diffcult because multiple clients may all modify object storage.

Ivanbeethoven:
# SingleFlight Correctness Note (SlayerFS)

This note summarizes the **correctness guarantees** and **limits** of SlayerFS `SingleFlight` (request coalescing), and explains why it is compatible with SlayerFS’ **close-to-open** consistency goal.

Key points:

- `SingleFlight` coalesces concurrent requests for the same key; it is a **performance primitive**, not a consistency mechanism.
- SlayerFS targets **close-to-open correctness**: after `close()` (or explicit `flush()` / `fsync()`) succeeds, subsequent `open()` observes the committed data/metadata.

## Where it lives

- Implementation: [src/chuck/singleflight.rs](../src/chuck/singleflight.rs)
- Primary use (coalesced full-block reads): [src/chuck/store.rs](../src/chuck/store.rs)
- close/flush path: [src/vfs/fs.rs](../src/vfs/fs.rs), [src/vfs/handles.rs](../src/vfs/handles.rs)
- reader invalidation / refresh: [src/vfs/io/reader.rs](../src/vfs/io/reader.rs)
- writer commit triggers invalidation: [src/vfs/io/writer.rs](../src/vfs/io/writer.rs)

## Model

For each key `K`, at any time there is either:

- an **executor** (the first caller that creates the in-flight entry and runs the operation), or
- one or more **waiters** (callers that observe the in-flight entry and wait for the executor’s result).

In SlayerFS, the typical key for object reads is:

- `BlockKey = (slice_id, block_index)`

This is important because `slice_id` is version-like (COW): different writes publish new `slice_id`s, producing new immutable object keys.

## Guarantees (SingleFlight-level)

### G1. At-most-one in-flight execution per key

For a given key, at most one task runs the underlying function `f()` at a time.

Reason: check/insert is serialized by `Mutex<HashMap<...>>` in [src/chuck/singleflight.rs](../src/chuck/singleflight.rs).

### G2. All waiters observe the same result as the executor

The executor computes exactly one `Result<Arc<V>, SharedError>` and broadcasts it. Each waiter receives a clone of the same `Arc`, i.e. shared ownership without copying `V`.

This is equivalent to JuiceFS’ “duplicate callers Acquire the returned Page” intent.

### G3. No “lost wakeup” between subscribe and send

The receiver subscribes while holding the map mutex; therefore a waiter cannot subscribe *after* the executor has already removed/sent for that in-flight entry.

If the executor crashes/aborts before sending, waiters fail with a channel-closed error (fail-fast; no infinite waiting).

### G4. In-flight entries are cleaned up

After the executor finishes, it removes the key from the map and then sends the final result.

## Non-goals (what SingleFlight does NOT guarantee)

- **No read/write consistency**: it does not linearize reads with concurrent writes, does not track inode/offset versions, and does not serialize cross-key operations.
- **No ordering across keys**.
- **No cancellation propagation** from waiters to the executor.

## Relationship to close-to-open correctness

SlayerFS’ close-to-open correctness is provided by the VFS + writer + reader pipeline, not by `SingleFlight`:

- `vfs.close()` flushes pending writes before returning success.
- Writes are COW-style: data is uploaded under a new `slice_id`, so old slice blocks are not overwritten.
- After metadata commit, the writer triggers reader cache invalidation/refresh for affected ranges.
- On handle close, the reader state for that handle is invalidated.

Why `SingleFlight` is safe under close-to-open:

- `SingleFlight` is keyed by `(slice_id, block_index)`; with COW, a committed write publishes new `slice_id`s, so subsequent opens/read paths naturally read **different keys**.
- Therefore, `SingleFlight` never merges reads across different versions of the same logical file range.

## Risks / audit checklist

- **Wrong key choice**: if a caller uses a key that can be overwritten in-place, `SingleFlight` may coalesce different versions and return stale/mixed data.
- **Executor abort**: waiters get an error (channel closed) and should rely on the caller’s retry policy.
- **Single-send assumption**: the implementation assumes one result per in-flight entry; capacity=1 is correct under that assumption.

## Tests

- Unit tests for basic coalescing/error behavior: [src/chuck/singleflight.rs](../src/chuck/singleflight.rs)

Optional (nice to have):

- A close-to-open regression test: writer handle writes + `close()`, then a fresh handle `open()` + read must observe new content (ensuring `SingleFlight` does not leak an older `(slice_id, block_index)` result).

Ivanbeethoven:
SLAYERFS_SINGLEFLIGHT_COMPARE=1 cargo run --example intelligent_read_demo
   Compiling slayerfs v0.1.0 (/home/luxian/code/rk8s/project/slayerfs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.32s
     Running `/home/luxian/code/rk8s/project/target/debug/examples/intelligent_read_demo`
🚀 SlayerFS smart read strategy demo
============================================================
📝 Creating 64MB test data...
✅ Test data written: 67108864 bytes

🔁 Comparison mode: SingleFlight OFF vs ON

📊 Smart read strategy test (SingleFlight=OFF):
   Threshold: 12.8MB (20% of 64MB block)
   Strategy: <= 12.8MB → range read | > 12.8MB → full read (no coalescing)

🔍 Scenario 1: small range read (10MB)
   ✓ Strategy: direct range read (get_object_range)
   ✓ Time: 6.298787ms (≈10.00 MB, 1587.61 MB/s)
   ✓ Data check: pass

🔍 Scenario 2: large range read (32MB)
   ✓ Strategy: full block read (no coalescing)
   ✓ Time: 49.212579ms (≈32.00 MB, 650.24 MB/s)
   ✓ Data check: pass

🔍 Scenario 3: concurrent large reads (no coalescing)
   ✓ Concurrent read 1 done
   ✓ Concurrent read 2 done
   ✓ Concurrent read 3 done
   ✓ Concurrent read 4 done
   ✓ Concurrent read 5 done
   ✓ Concurrent read 6 done
   ✓ Concurrent read 7 done
   ✓ Concurrent read 8 done
   ✓ Concurrent read 9 done
   ✓ Concurrent read 10 done
   ✓ Total time: 368.032125ms (total ≈320.00 MB, downstream IO ≈ concurrency count, throughput ≈869.49 MB/s)

📊 Smart read strategy test (SingleFlight=ON):
   Threshold: 12.8MB (20% of 64MB block)
   Strategy: <= 12.8MB → range read | > 12.8MB → full read + SingleFlight

🔍 Scenario 1: small range read (10MB)
   ✓ Strategy: direct range read (get_object_range)
   ✓ Time: 6.089499ms (≈10.00 MB, 1642.17 MB/s)
   ✓ Data check: pass

🔍 Scenario 2: large range read (32MB)
   ✓ Strategy: full block read + SingleFlight coalescing
   ✓ Time: 52.141367ms (≈32.00 MB, 613.72 MB/s)
   ✓ Data check: pass

🔍 Scenario 3: concurrent large reads (show SingleFlight effect)
   ✓ Concurrent read 1 done
   ✓ Concurrent read 2 done
   ✓ Concurrent read 3 done
   ✓ Concurrent read 4 done
   ✓ Concurrent read 5 done
   ✓ Concurrent read 6 done
   ✓ Concurrent read 7 done
   ✓ Concurrent read 8 done
   ✓ Concurrent read 9 done
   ✓ Concurrent read 10 done
   ✓ Total time: 81.305334ms (total ≈320.00 MB, downstream IO ≈ 1 after coalescing, throughput ≈3935.78 MB/s)

project/slayerfs/src/chuck/singleflight.rs
Tyooughtul: perhaps using dashmap is a better choice

project/slayerfs/src/chuck/singleflight.rs
```diff
@@ -109,11 +113,16 @@ where
...
 
-        // If we're a waiter, wait for the result
+        // If we're a waiter, wait for the result. Safe from races: subscription is created
+        // while holding the mutex so even if the executor finishes immediately after unlock,
+        // the broadcasted value will still be delivered to this receiver.
         if let Some(ref mut rx) = rx {
             return match rx.recv().await {
                 Ok(result) => result,
-                Err(_) => Err(Arc::new(anyhow::anyhow!("SingleFlight: channel closed"))),
+                Err(_) => Err(Arc::new(anyhow::anyhow!(
```
Tyooughtul: Because broadcast is used instead of oneshot and the channel is only set to 1, I think there is still a risk under concurrent conditions. So I think that to "Err (broadcast: : RecvError: : Closed) = > Err (...). , 
Err (broadcast: : RecvError: : Lagged (_)) = > "would be better

project/slayerfs/src/chuck/singleflight.rs
Tyooughtul: if f paniced or killed,
```rust
{
            let mut guard = self.in_flight.lock().unwrap();
            if let Some(in_flight) = guard.remove(&key) {
            // Ignore send errors - no receivers means no one is waiting
            let _ = in_flight.tx.send(shared_result.clone());
    }
``` 
won't be executed, the key will be left in the in_flight ?

genedna (commented):
## Code Review: Slayerfs SingleFlight Improvements

Overall this PR makes solid improvements to the SingleFlight coalescing mechanism and adds much-needed tests for concurrent behavior. The core motivation — cloning `ObjectClient` to avoid lifetime issues with the async closure — is sound. However, there are several issues worth addressing before merge.

---

### 👍 What's Good

- **Concurrent coalescing test** (`test_concurrent_requests_coalesced` in store.rs and `test_concurrent_requests_executor_fails` in singleflight.rs) — great additions that actually validate the SingleFlight contract under concurrent load.
- **`BlockStoreConfig::validate()`** — good defensive programming to catch invalid configurations early.
- **`range_size_threshold()` helper** — reduces duplication and centralizes the threshold calculation.
- **Improved error message** in SingleFlight channel-closed path now includes the key for debuggability.
- **Thread-safety documentation** on `SingleFlight` and `read_flight` field.
- **Tracing cleanup** — consolidating the `read_len` recording to a single site.

---

### ⚠️ Issues & Suggestions

See inline comments for specific code-level feedback. Summary of key concerns:

1. **Breaking API change**: Adding `Clone` bound to `ObjectBackend` in `ObjectClient` and `ObjectBlockStore` is a **breaking change** for downstream consumers. Any existing `ObjectBackend` impl that isn't `Clone` will fail to compile. Consider whether this can be scoped more narrowly (e.g., only where SingleFlight is used).

2. **`Debug` bound on SingleFlight keys**: The new `std::fmt::Debug` bound on `K` is also a breaking change for existing `SingleFlight` users with non-Debug key types.

3. **Demo allocates 64MB on the stack-adjacent heap** — while not a bug, the demo is now significantly more resource-hungry and should be documented/noted.

4. **Threshold semantics changed silently** — `<` to `<=` changes the boundary behavior. This is fine but should be called out in commit messages.

5. **Missing `put_object_vectored` in MockBackend** — the mock doesn't override the default vectored implementation, which may mask issues in tests.

project/slayerfs/src/cadapter/client.rs
```diff
@@ -30,11 +34,11 @@ pub trait ObjectBackend: Send + Sync {
 }
 
 #[derive(Clone)]
-pub struct ObjectClient<B: ObjectBackend> {
+pub struct ObjectClient<B: ObjectBackend + Clone> {
```
genedna: **Breaking API change**: Adding `Clone` bound here tightens the contract for `ObjectClient<B>`. Every callsite that constructs an `ObjectClient` now requires `B: ObjectBackend + Clone`.

Consider keeping this struct generic over just `ObjectBackend` and only adding the `Clone` bound where it's actually needed — i.e., on the `ObjectBlockStore` impl that uses SingleFlight. You could use a `where B: Clone` clause on specific methods rather than on the struct definition itself:

```rust
pub struct ObjectClient<B: ObjectBackend> { ... }  // no Clone bound

// Only where cloning is needed:
impl<B: ObjectBackend + Clone> ObjectBlockStore<B> { ... }
```

This preserves backward compatibility for consumers that don't need SingleFlight.

project/slayerfs/src/chuck/singleflight.rs
```diff
@@ -48,7 +52,7 @@ pub struct SingleFlight<K, V> {
 
 impl<K, V> Default for SingleFlight<K, V>
 where
-    K: Hash + Eq + Clone,
+    K: Hash + Eq + Clone + std::fmt::Debug,
```
genedna: **Breaking change**: Adding `Debug` bound on `K` is a trait contract expansion. While `BlockKey = (u64, u32)` trivially satisfies `Debug`, this constrains all future uses of `SingleFlight`.

Since `Debug` is only used in the error message at one callsite (`"SingleFlight: channel closed for key: {:?}"`), consider making the `Debug` bound conditional only on the `execute` method rather than on the struct/Default impl. Alternatively, if `SingleFlight` is internal-only, this is acceptable — but document that intent.

project/slayerfs/src/chuck/store.rs
```diff
@@ -317,7 +338,7 @@ impl<B: ObjectBackend + Send + Sync> BlockStore for ObjectBlockStore<B> {
 
         // Use SingleFlight to coalesce concurrent reads to the same block.
         // We read the entire block and then extract the requested range.
-        let client = &self.client;
+        let client: ObjectClient<B> = self.client.clone();
```
genedna: **Performance concern**: Cloning `ObjectClient<B>` on every large read could be expensive depending on the backend implementation. If `B` contains connection pools, TLS state, or other heavyweight resources, each clone creates a copy.

Consider wrapping `ObjectClient` in `Arc` at the `ObjectBlockStore` level so the clone here is just an `Arc` increment:

```rust
pub struct ObjectBlockStore<B: ObjectBackend> {
    client: Arc<ObjectClient<B>>,  // cheap to clone
    ...
}
```

This also removes the need for the `Clone` bound on `ObjectBackend` entirely — you'd just clone the `Arc`.

project/slayerfs/src/chuck/store.rs
```diff
@@ -152,7 +153,23 @@ impl Default for BlockStoreConfig {
...
+impl BlockStoreConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        if self.block_size == 0 {
+            anyhow::bail!("block_size must be greater than 0");
+        }
+        if !(0.0..=1.0).contains(&self.range_read_threshold) {
+            anyhow::bail!("range_read_threshold must be between 0.0 and 1.0");
+        }
+        Ok(())
+    }
```
genedna, line 165: **Nit**: `validate()` on the default config should never fail, but if it does, `expect()` will panic with a non-actionable message. Since `new()` doesn't return `Result`, this is a design smell — consider either:
1. Making `new()` return `Result<Self>` (consistent with `new_with_configs`)
2. Using `debug_assert!` if this is truly "cannot happen"
3. Using `const` validation at compile time if the default values are known

project/slayerfs/src/chuck/store.rs
```diff
@@ -152,7 +153,23 @@ impl Default for BlockStoreConfig {
...
+
+    fn range_size_threshold(&self) -> usize {
+        (self.block_size as f32 * self.range_read_threshold) as usize
+    }
+}
```
genedna, line 170: **Suggestion**: The `range_read_threshold` helper casts through `f32`, which loses precision for large block sizes. For a 64MB block (`67_108_864`) with threshold 0.2:

```
(67_108_864 as f32 * 0.2) as usize = 13_421_772
```

This is fine for practical purposes, but if exact thresholds matter, consider using integer arithmetic:
```rust
fn range_size_threshold(&self) -> usize {
    (self.block_size as u64 * (self.range_read_threshold * 1000.0) as u64 / 1000) as usize
}
```

project/slayerfs/src/chuck/store.rs
```diff
@@ -291,13 +311,14 @@ impl<B: ObjectBackend + Send + Sync> BlockStore for ObjectBlockStore<B> {
     // Caller is responsible for zero-filling buf; this method only overwrites existing bytes.
     async fn read_range(&self, key: BlockKey, offset: u64, buf: &mut [u8]) -> anyhow::Result<()> {
         let len = buf.len();
-        let range_size_threshold =
-            (self.config.block_size as f32 * self.config.range_read_threshold) as usize;
+        let range_size_threshold = self.config.range_size_threshold();
+
+        // Boundary: len == threshold still uses direct range read; threshold is floor-casted usize.
```
genedna, line 316: **Correctness**: The `<=` change (from `<`) means that reads exactly at the threshold now take the direct range path instead of the coalesced path. This is a behavioral change that should be documented in the commit message.

Also, the inline comment says "threshold is floor-casted usize" — this means for `block_size=64MB, threshold=0.2`, the boundary is at `13_421_772` bytes (≈12.8MB). Reads of exactly this size go to range read. Verify this is the intended behavior, especially for tests that may rely on the old `<` boundary.

project/slayerfs/src/chuck/singleflight.rs
```diff
@@ -109,11 +113,16 @@ where
             }
         };
 
-        // If we're a waiter, wait for the result
+        // If we're a waiter, wait for the result. Safe from races: subscription is created
+        // while holding the mutex so even if the executor finishes immediately after unlock,
+        // the broadcasted value will still be delivered to this receiver.
         if let Some(ref mut rx) = rx {
             return match rx.recv().await {
```
genedna: Good comment about the broadcast subscription race safety. This is a subtle correctness point — the `subscribe()` call happens while holding the mutex, so even if the executor completes immediately after we drop the lock, the `Receiver` was already registered and will get the value.

**However**, there's a subtle issue: `broadcast::channel(1)` has a capacity of 1. If the executor finishes and calls `send()` before the waiter calls `recv()`, the value is stored in the channel buffer (capacity 1). But if a second value were ever sent (which doesn't happen here), it could overflow. The current code is correct, but the capacity of 1 is load-bearing — worth a comment.

project/slayerfs/src/chuck/store.rs
```diff
@@ -486,6 +508,8 @@ mod tests {
             }
 
             async fn get_object(&self, key: &str) -> anyhow::Result<Option<Vec<u8>>> {
+                // Simulate some latency so SingleFlight can coalesce concurrent requests
```
genedna, line 513: **Testing**: The `MockBackend::get_object` now has a `sleep(Duration::from_millis(10))` to simulate latency. This is necessary for the concurrent coalescing test but makes all other tests that use this mock (including the small-range test above) slightly slower.

Consider making the latency configurable (e.g., via a field on `MockBackend`) so it's only enabled for tests that need it.

project/slayerfs/examples/intelligent_read_demo.rs
```diff
@@ -29,28 +29,34 @@ async fn main() -> Result<()> {
...
     )?);
 
-    // 创建测试数据 (4MB 的测试块)
-    println!("📝 创建 4MB 测试数据...");
-    let test_data: Vec<u8> = (0..4_194_304).map(|i| (i % 256) as u8).collect();
+    // 创建测试数据 (64MB 的测试块)
+    println!("📝 创建 64MB 测试数据...");
+    let test_data: Vec<u8> = (0..67_108_864).map(|i| (i % 256) as u8).collect();
 
     // 写入测试数据
```
genedna: **Resource concern**: This demo now allocates 64MB of test data, plus 10MB, 32MB, and 10×32MB buffers during concurrent tests. That's ~450MB total heap allocation. Consider adding a note at the top of the demo about memory requirements, or providing a CLI flag to control block size for resource-constrained environments.

project/slayerfs/src/daemon/worker.rs
```diff
@@ -35,14 +35,14 @@ impl Default for GcConfig {
 }
 
 #[allow(dead_code)]
-pub(crate) struct MarkBasedGarbageCollector<B: ObjectBackend> {
+pub(crate) struct MarkBasedGarbageCollector<B: ObjectBackend + Clone> {
```
genedna, line 38: The `Clone` bound propagation here (`ObjectBackend + Clone`) is expected given the upstream change, but it widens the contract for `MarkBasedGarbageCollector` which doesn't actually need `Clone` for its own operations. This is collateral damage from putting `Clone` on `ObjectClient`.

This reinforces the suggestion to use `Arc<ObjectClient<B>>` instead — it would avoid needing `Clone` bounds in places like the GC that don't clone the client.

PR #458, opened 2026-02-13
slayerfs: flatten CacheSlice pages in vfs

Ivanbeethoven:


mon3stera:
Why do we have to flatten the pages?

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -18,7 +18,15 @@ pub(crate) struct CacheSlice {
     config: Arc<WriteConfig>,
     len: u64,
     alloc_bytes: u64,
-    pages: Vec<Vec<Page>>,
+    pages: Vec<Option<Page>>, // 优化为单层Vec
```
genedna: **[Perf - Medium]** Pre-allocation increases baseline memory.

The old structure used `vec![Vec::new(); blocks]` — each block started as an empty `Vec` with zero heap allocation until pages were actually written. The new `vec![None; total_pages]` pre-allocates `blocks * pages_per_block` `Option<Page>` slots upfront.

For the test config (chunk=16KB, block=4KB, page=1KB): old = 4 empty Vecs (~96 bytes), new = 16 `Option<Page>` slots. This is fine for small configs, but for large chunk sizes (e.g. 64MB chunks with 4KB pages), this could mean pre-allocating millions of `None` slots.

Consider whether lazy growth or a `HashMap<usize, Page>` might be more appropriate if chunk sizes can be large, or add a comment documenting the assumption about bounded chunk sizes.

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -77,17 +86,18 @@ impl CacheSlice {
...
-            let span = block_span.split_into::<PageTag>(block_size as u64, page_size as u64, true);
-
-            for page_span in span {
-                let page = &mut self.pages[block_span.index.as_usize()][page_span.index.as_usize()];
+        for block_span in span.split_into::<BlockTag>(chunk_size, block_size, true) {
+            let page_spans = block_span.split_into::<PageTag>(block_size, page_size, true);
+            for page_span in page_spans {
+                let block_idx = block_span.index.as_usize();
+                let page_idx = page_span.index.as_usize();
+                let flat_idx = self.flat_index(block_idx, page_idx, pages_per_block);
```
genedna: **[Correctness - Medium]** `write_at` behavioral change: now silently allocates pages.

The old `write_at` accessed pages directly with `&mut self.pages[block][page]`, which would panic if the page didn't exist. The new code uses `ensure_page_mut_by_flat()`, which allocates a new page if `None`.

This is a **behavioral change**, not just a refactor. In the old code, `write_at` on an unallocated page would panic (indicating a bug in the caller). Now it silently allocates, which could mask bugs where `write_at` is called on regions that were never appended to.

Is this intentional? If so, please document this design choice. If not, consider checking `is_some()` first and returning an error for writes to unallocated pages.

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -195,50 +213,90 @@ impl CacheSlice {
...
+        (flat_idx, within_page)
+    }
+
+    fn pages_per_block(&self) -> usize {
+        let page_size = self.config.page_size as usize;
+        let block_size = self.config.layout.block_size as usize;
+        block_size.div_ceil(page_size)
+    }
+
+    fn flat_index(&self, block_idx: usize, page_idx: usize, pages_per_block: usize) -> usize {
```
genedna, line 292: **[Bug - High]** `flat_index_checked` is missing `page_idx` bounds validation.

This only validates the final flat index is in range, but does NOT validate `page_idx < pages_per_block`. Example:
- `flat_index(block_idx=0, page_idx=5, pages_per_block=4)` → `flat_idx = 5`
- This is a valid flat index, but it **silently aliases into block 1, page 1** instead of returning `None`

This can lead to incorrect reads/writes across block boundaries.

**Suggested fix:**
```rust
fn flat_index_checked(&self, block_idx: usize, page_idx: usize, pages_per_block: usize) -> Option<usize> {
    if page_idx >= pages_per_block {
        return None;
    }
    let flat_idx = self.flat_index(block_idx, page_idx, pages_per_block);
    if flat_idx < self.pages.len() {
        Some(flat_idx)
    } else {
        None
    }
}
```

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -195,50 +213,90 @@ impl CacheSlice {
...
+        page_idx: usize,
+        pages_per_block: usize,
+    ) -> Option<usize> {
+        let flat_idx = self.flat_index(block_idx, page_idx, pages_per_block);
+        if flat_idx < self.pages.len() {
+            Some(flat_idx)
+        } else {
+            None
+        }
+    }
```
genedna: **[Code Quality - Medium]** `page()` and `page_mut()` are dead code — not called anywhere in this PR or the existing codebase.

If intended for future use:
1. Add `#[allow(dead_code)]` with a comment explaining the planned use case
2. Add unit tests for them now

If not needed yet, remove them (YAGNI principle) and add them when actually needed.

Also note: `page_mut()` returns `Option<&mut Page>` but doesn't allocate — this is different from `ensure_page_mut_by_flat()` which allocates on demand. The naming distinction should be documented.

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -195,50 +213,90 @@ impl CacheSlice {
...
-    /// The returned slot belongs to a single page and never overlaps existing writes.
-    fn next_write_slot(&mut self) -> (usize, usize, usize) {
+    /// 获取下一个可写slot（扁平索引，页内偏移）
+    fn next_write_slot_flat(&mut self) -> (usize, usize) {
         let page_size = self.config.page_size as usize;
         let block_size = self.config.layout.block_size as usize;
-
+        let pages_per_block = self.pages_per_block();
         let total = self.len.as_usize();
         let block_index = total / block_size;
```
genedna: **[Documentation - Low]** Mixed language comments.

```rust
/// 获取下一个可写slot（扁平索引，页内偏移）
```
and line 21:
```rust
pages: Vec<Option<Page>>, // 优化为单层Vec
```

The rest of the codebase uses English doc comments. For consistency:
```rust
/// Acquire the next writable slot as (flat_index, offset_within_page).
fn next_write_slot_flat(&mut self) -> (usize, usize) {
```
```rust
pages: Vec<Option<Page>>, // Flattened to a single Vec for simpler indexing
```

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -195,50 +213,90 @@ impl CacheSlice {
...
         let within_block = total % block_size;
         let page_index = within_block / page_size;
         let within_page = within_block % page_size;
+        let flat_idx = self.flat_index(block_index, page_index, pages_per_block);
+        (flat_idx, within_page)
```
genedna, line 285: **[Perf - Low]** `pages_per_block()` is recomputed via division on every call, and it's called in nearly every method.

Since `config` is immutable after construction, consider caching this as a field:
```rust
pub(crate) struct CacheSlice {
    config: Arc<WriteConfig>,
    len: u64,
    alloc_bytes: u64,
    pages_per_block: usize,  // computed once in new()
    pages: Vec<Option<Page>>,
}
```

This would also eliminate the need to pass `pages_per_block` as a parameter to `flat_index()` and friends, simplifying the internal API.

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -195,50 +213,90 @@ impl CacheSlice {
...
+    }
+
+    fn pages_per_block(&self) -> usize {
+        let page_size = self.config.page_size as usize;
```
genedna: **[Code Quality - Low]** `flat_index` takes `pages_per_block` as a parameter even though it has `&self` access.

This creates an inconsistency — callers must pass the correct value manually. If the value were cached as a field, this parameter would be unnecessary.

At minimum, add a debug assertion:
```rust
fn flat_index(&self, block_idx: usize, page_idx: usize, pages_per_block: usize) -> usize {
    debug_assert_eq!(pages_per_block, self.pages_per_block());
    block_idx * pages_per_block + page_idx
}
```

project/slayerfs/src/vfs/cache/page.rs
```diff
@@ -135,43 +143,53 @@ impl CacheSlice {
...
-        for block in self.pages[start..end].iter_mut() {
-            for page in block.iter_mut() {
-                page.freeze();
-            }
+        let pages_per_block = self.pages_per_block();
+        let start_idx = start * pages_per_block;
+        let end_idx = end * pages_per_block;
+        for page in self.pages[start_idx..end_idx].iter_mut().flatten() {
+            page.freeze();
         }
```
genedna, line 192: **[Testing - Medium]** Missing test coverage for new public API.

The following new methods lack tests:
- `stats()` — verify `pages_total`, `pages_used`, `len`, `alloc_bytes` are correct
- `page()` / `page_mut()` — verify access and out-of-bounds returns `None`
- `flat_index_checked()` — especially important given the bounds bug noted above

Suggested:
```rust
#[test]
fn test_stats_after_append() {
    let mut slice = CacheSlice::new(config());
    slice.append(&patterned(512, 1)).unwrap();
    let stats = slice.stats();
    assert_eq!(stats.len, 512);
    assert_eq!(stats.pages_used, 1);
    assert!(stats.pages_total > 0);
}

#[test]
fn test_page_out_of_bounds() {
    let slice = CacheSlice::new(config());
    assert!(slice.page(100, 0).is_none());
}
```

PR #465, opened 2026-02-15
fix(distribution): make compose deployment work with workspace build …

Lingbou:
- build distribution from workspace root with cargo build -p distribution
- switch compose build context to project root and use distribution/Dockerfile
- use db:5432 in DATABASE_URL instead of localhost for container networking
- wait for postgres healthcheck before starting distribution


genedna:
@codex review

genedna (approved):

PR #467, opened 2026-02-16
feat(rkforge): align build -t semantics with image refs and tag-aware…

Lingbou:
- change build -t/--tag to parse OCI image references and support multiple tags
- derive output OCI layout directory from the first tag safely (instead of raw tag path)
- propagate tag ref names through executor and write them into OCI index annotations
- support multiple ref names in index manifests (instead of hardcoded latest)
- make push tag-aware by digest/ref mapping:
  - explicit tag: push only matching manifest(s), error if tag not found in index
  - implicit tag: keep existing multi-ref expansion behavior
- add/adjust tests for tag parsing, explicit-tag detection, ref dedup, and push filtering

### DIFF
--- a/project/rkforge/README.md
+++ b/project/rkforge/README.md
@@ -120,6 +120,23 @@ Alternatively, use `--libfuse` option to avoid sudo (uses userspace FUSE):
 
 Here the last optional `.` specifies the build context, which by default is `.`.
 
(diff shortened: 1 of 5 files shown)


Lingbou:
#466

genedna:
@codex review

genedna:
@claude

PR #472, opened 2026-02-22
feat:Xline-client is ported to use the new Channel

briteny-pwn:


yyjeqhc (changes requested):
RPC calls should use encoded method identifiers instead of full string paths to reduce overhead.

PR #477, opened 2026-02-24
feat(libcsi): volume persistence, recovery, and idempotency improvements

Ricky-Daxia:
## Summary

- **Volume state persistence**: each volume now writes a `<vol-id>.meta.json` sidecar under `object_root` at creation time. A new `recover()` method scans these files on startup to rebuild `volumes` / `volume_meta` / `volume_names` after a process restart, eliminating the previous in-memory-only state that was lost on every restart.

- **`create_volume` idempotency**: a new `volume_names: DashMap<String, VolumeId>` map and the `PARAM_CSI_NAME` constant (embedded in `Volume::parameters`) let the backend detect and return existing volumes for repeated calls with the same name, conforming to the CSI spec's idempotency requirement.

- **`delete_volume` ordering**: on-disk data (volume directory + `.meta.json`) is removed *before* in-memory state is cleared. This means a crash mid-delete leaves state intact and the operation can be safely retried.

- **`unstage_volume` staging-dir cleanup**: the staging directory is removed after the FUSE unmount so that a subsequent `stage_volume` call can use directory existence as an unambiguous "is mounted" signal.

- **`publish_volume` / `unpublish_volume` idempotency**: replaced directory-existence checks with `is_mountpoint()` (reads `/proc/self/mounts`). This prevents spurious `EBUSY` on a double publish and `EINVAL` on a double unpublish.

- **`CsiClient::connect` server name**: the TLS SNI string is now a caller-supplied `server_name` parameter instead of the hardcoded `"libcsi"` literal, allowing the QUIC client to authenticate against certificates with arbitrary SANs issued by `libvault`.

- **New unit tests**: `create_volume_idempotent_by_name` and `recover_restores_state` cover the two new correctness properties.

- **`docs/design.md`**: updated all relevant code snippets and the roadmap table to reflect the implementation.

## Test plan

- [ ] `cargo test --lib -p libcsi` — all unit tests pass, including the two new tests
- [ ] Manual end-to-end: create a volume, kill the process, restart and call `recover()`, confirm the volume is still accessible
- [ ] Verify `CsiClient::connect` with a real `libvault`-issued certificate using the correct `server_name`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/project/libcsi/docs/design.md
+++ b/project/libcsi/docs/design.md
@@ -415,7 +415,8 @@ impl CsiClient {
         endpoint.set_default_client_config(quinn::ClientConfig::new(Arc::new(
             quinn::crypto::rustls::QuicClientConfig::try_from(tls_config)?
         )));
-        let connection = endpoint.connect(addr, "libcsi")?.await?;
+        // server_name must match a SAN in the server certificate (issued by libvault).
+        let connection = endpoint.connect(addr, server_name)?.await?;
         Ok(Self { connection })
     }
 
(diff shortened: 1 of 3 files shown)


Ricky-Daxia:
Closing in favor of a more complete PR that covers the full libcsi addition.

PR #497, opened 2026-03-03
feat(rkforge/build): add --secret and --ssh support for RUN

Lingbou:
- add BuildSecret/BuildSshAgent models and wire them through build context/executor/task
- add --secret and --ssh CLI parsing with validation (secret id traversal guard, ssh unix-socket check)
- pass secrets/ssh into exec-internal via base64 JSON arguments
- prepare runtime mounts for /run/secrets and /run/rkforge/ssh, with tmpfs + bind mounts and cleanup
- set default SSH_AUTH_SOCK for RUN when ssh agents are configured
- add/adjust tests for secret/ssh parsing and related edge cases


Lingbou:
#466

PR #500, opened 2026-03-04
fix(rks):add network for rks without rkl

tcytan:


genedna (changes requested):
## Code Review: fix(rks): add network for rks without rkl

Overall this PR makes good progress toward enabling standalone RKS networking without RKL. The `NetworkConfigMessage` simplification and dependency upgrades are solid. However, there are several issues — one critical — that should be addressed before merging.

---

### Summary of Changes
- **New** `local_node.rs` module (320 lines) for bootstrapping local RKS node networking
- **Simplified** `NetworkConfigMessage::SubnetConfig` to carry a raw `subnet_env: String`
- **Moved** subnet-env parsing from `client.rs` to `receiver.rs` (and duplicated in `local_node.rs`)
- **Upgraded** `netlink-packet-route` 0.22→0.25.1, `rtnetlink` 0.16→0.18.1
- **Removed** `qlean` dependency from `libvault`; bumped `libvault` 0.1.0→0.2.2
- **Added** Windows `rustc_flags` for `rkforge` and `rkl` BUCK targets
- **Increased** QUIC message buffer 4096→8192 bytes

---

### 🔴 Critical Issues

#### 1. Stale routes never cleaned up in `apply_routes()` (`local_node.rs:240-278`)

`apply_routes()` only **adds** routes but never **removes** stale ones. When a worker lease is deleted or its subnet changes, the old route remains in the kernel routing table, potentially forwarding traffic to a dead gateway. This is a data-plane correctness bug.

The `RouteManager` already has `delete_route()` and `cleanup_routes()` methods. The fix should compute a diff between old and new route sets, deleting entries no longer present:

```rust
async fn apply_routes(route_manager: Arc<Mutex<RouteManager>>, routes: Vec<Route>) {
    let mut manager = route_manager.lock().await;
    // Remove stale routes first
    let old_v4 = manager.get_routes().to_vec();
    for old in &old_v4 {
        if !routes.iter().any(|r| route::route_equal(r, old)) {
            let _ = manager.delete_route(old).await;
        }
    }
    // Then add new/updated routes...
}
```

---

### 🟡 Important Issues

#### 2. Blocking bootstrap delays QUIC server startup (`mod.rs:118-133`)

`start_background_tasks()` was previously `fn` (sync) and is now `async`. The `bootstrap().await` call blocks the QUIC server from starting until the full bootstrap completes (network config fetch → ext interface discovery → lease acquisition). If any step hangs, the RKS node cannot accept incoming connections.

**Recommendation:** Spawn bootstrap as a background task using `tokio::spawn()` to keep server startup non-blocking:
```rust
let shared_clone = self.shared.clone();
let addr_clone = self.addr.clone();
tokio::spawn(async move {
    if let Err(e) = local_node::bootstrap(shared_clone, &addr_clone).await {
        warn!("Failed to bootstrap local rks node networking: {e:#}");
    }
});
```

#### 3. Duplicated subnet-env parsing logic (`local_node.rs:170-215` vs `receiver.rs:88-130`)

The same `KEY=VALUE` parsing exists in both files with **subtle differences**:
- `local_node.rs` handles `RKL_IPV6_SUBNET` ✅
- `receiver.rs` does **not** handle `RKL_IPV6_SUBNET` ❌
- `receiver.rs` handles `RKL_NETWORK`, `local_node.rs` does not

This drift will worsen over time. Extract a shared `SubnetEnvConfig::parse(env_str)` function.

#### 4. IPv6 subnet not parsed in `receiver.rs` (line 124)

`receiver.rs` hardcodes `ipv6_subnet` to `None`:
```rust
.handle_subnet_config(&network_config, ip_masq, subnet, None, mtu)
```
While `local_node.rs` correctly parses `RKL_IPV6_SUBNET`. This means IPv6 networking silently fails for workers connecting via RKL.

---

### 🔵 Minor Issues

#### 5. Fixed buffer size may still be insufficient (`client.rs:181`)
Buffer bumped 4096→8192, but payloads with many routes could exceed this. Consider reading in a loop or using `read_to_end()` with a max bound. Truncated reads cause silent JSON parse failures.

#### 6. No unit tests for `local_node.rs`
This 320-line file has zero tests. Pure-logic functions like `resolve_local_node_id()`, `sanitize_node_suffix()`, and the subnet-env parser are easily unit-testable without mocking.

#### 7. Unrelated Windows rustc_flags changes (`rkforge/BUCK`, `rkl/BUCK`)
The Windows build flag additions appear unrelated to networking. Consider splitting into a separate PR for cleaner git history.

---

### ✅ What Looks Good
- The `NetworkConfigMessage` simplification is a clean improvement — passing the raw string and parsing on the receiver side is the right direction
- Dependency upgrades (netlink-packet-route, rtnetlink) are well-scoped
- Logging in `local_node.rs` is thorough with good use of log targets
- Error handling with `anyhow::Context` provides good error chains
- The `sanitize_node_suffix()` function handles edge cases well

project/rks/src/node/local_node.rs
```diff
@@ -0,0 +1,320 @@
...
+                if let Err(e) = manager.add_v6_route(&route).await {
+                    error!(
+                        target: "rks::node::local",
+                        "failed to add IPv6 route {route:?}: {e}"
+                    );
+                }
+            }
+            None => {
+                warn!(
+                    target: "rks::node::local",
```
genedna, line 269: **🔴 Bug: Stale routes never cleaned up**

`apply_routes()` only adds routes but never removes stale ones. When a worker lease is deleted or its subnet changes, the old route stays in the kernel routing table, forwarding traffic to a dead gateway.

`RouteManager` already has `delete_route()`. Consider computing a diff:

```rust
// Before adding, remove routes no longer in the new set
let old_v4 = manager.get_routes().to_vec();
for old in &old_v4 {
    if !routes.iter().any(|r| route::route_equal(r, old)) {
        let _ = manager.delete_route(old).await;
    }
}
```

project/rks/src/node/mod.rs
```diff
@@ -127,6 +128,12 @@ impl RksNode {
         );
         info!("Heartbeat monitor started");
 
+        if let Err(e) = local_node::bootstrap(self.shared.clone(), &self.addr).await {
+            warn!("Failed to bootstrap local rks node networking: {e:#}");
+        } else {
```
genedna: **🟡 Blocking bootstrap delays QUIC server startup**

`bootstrap().await` runs before `QUICServer::new()`. If lease acquisition or interface discovery hangs, the server never starts. Consider wrapping this in `tokio::spawn()` to keep it non-blocking.

project/rks/src/node/local_node.rs
```diff
@@ -0,0 +1,320 @@
...
+        {
+            warn!(
+                target: "rks::node::local",
+                "lease completion loop ended for node {}: {e:#}",
+                node_id
+            );
+        }
+    });
+}
+
```
genedna, line 194: **🟡 Duplicated subnet-env parsing**

This same KEY=VALUE parsing exists in `receiver.rs:88-130` with subtle differences (this file handles `RKL_IPV6_SUBNET`, `receiver.rs` does not). Extract a shared parser to avoid drift.

project/rkl/src/network/receiver.rs
```diff
@@ -81,15 +85,45 @@ impl NetworkReceiver {
...
+                    ipv6_subnet_max: None,
+                    subnet_len: 24,
+                    ipv6_subnet_len: 64,
+                    backend_type: std::env::var("BACKEND_TYPE")
+                        .unwrap_or_else(|_| "hostgw".to_string()),
+                    backend: None,
+                };
+
+                validate_network_config(&mut network_config)?;
+
```
genedna, line 124: **🟡 IPv6 subnet not parsed here**

`local_node.rs` correctly parses `RKL_IPV6_SUBNET`, but this code hardcodes `None`. IPv6 networking will silently fail for RKL workers. Add `RKL_IPV6_SUBNET` parsing to match `local_node.rs`.

project/rkl/src/daemon/client.rs
```diff
@@ -182,7 +178,7 @@ pub async fn run_once(
     loop {
         match client.accept_uni().await {
             Ok(mut recv) => {
-                let mut buf = vec![0u8; 4096];
+                let mut buf = vec![0u8; 8192];
```
genedna, line 181: **🔵 Fixed buffer may be insufficient**

Bumped 4096→8192, but large payloads could still truncate. Consider `recv.read_to_end()` or looped reads with a max bound to avoid silent JSON parse failures.

PR #502, opened 2026-03-05
fix(libfuse-fs): improve async file-lock semantics and unify fallback behavior

Ivanbeethoven:
1. passthrough: implement getlk/setlk via fcntl, supporting both blocking and non-blocking locks
2. overlayfs/unionfs: forward getlk/setlk to the underlying real handle
3. add compatibility fallback for underlying ENOSYS: getlk returns F_UNLCK, setlk degrades to success
4. downgrade ESTALE in open_by_handle_at to debug logging to reduce false alarms

### DIFF
--- a/project/libfuse-fs/src/overlayfs/async_io.rs
+++ b/project/libfuse-fs/src/overlayfs/async_io.rs
@@ -846,32 +846,102 @@ impl Filesystem for OverlayFs {
     #[allow(clippy::too_many_arguments)]
     async fn getlk(
         &self,
-        _req: Request,
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
+        fh: u64,
+        lock_owner: u64,
+        start: u64,
+        end: u64,
+        r#type: u32,
+        pid: u32,
     ) -> Result<ReplyLock> {
-        Err(libc::ENOSYS.into())
+        if !self.no_open.load(Ordering::Relaxed) {
+            let handles = self.handles.lock().await;
+            if let Some(hd) = handles.get(&fh)
+                && let Some(ref rh) = hd.real_handle
+            {
+                match rh
+                    .layer
+                    .getlk(
+                        req,
+                        rh.inode,
+                        rh.handle.load(Ordering::Relaxed),
+                        lock_owner,
+                        start,
+                        end,
+                        r#type,
+                        pid,
+                    )
+                    .await
+                {
+                    Ok(reply) => return Ok(reply),
+                    Err(e) => {
+                        // If underlying layer doesn't support locking, fall through to fallback
+                        let errno: i32 = e.into();
+                        if errno != libc::ENOSYS {
+                            return Err(errno.into());
+                        }
+                    }
+                }
+            }
+        }
+
+        // Fallback: report no lock conflict
+        Ok(ReplyLock {
+            start: 0,
+            end: 0,
+            r#type: libc::F_UNLCK as u32,
+            pid: 0,
+        })
     }
 
     #[allow(clippy::too_many_arguments)]
     async fn setlk(
         &self,
-        _req: Request,
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
-        _block: bool,
+        fh: u64,
+        lock_owner: u64,
+        start: u64,
+        end: u64,
+        r#type: u32,
+        pid: u32,
+        block: bool,
     ) -> Result<()> {
-        Err(libc::ENOSYS.into())
+        if !self.no_open.load(Ordering::Relaxed) {
+            let handles = self.handles.lock().await;
+            if let Some(hd) = handles.get(&fh)
+                && let Some(ref rh) = hd.real_handle
+            {
+                match rh
+                    .layer
+                    .setlk(
+                        req,
+                        rh.inode,
+                        rh.handle.load(Ordering::Relaxed),
+                        lock_owner,
+                        start,
+                        end,
+                        r#type,
+                        pid,
+                        block,
+                    )
+                    .await
+                {
+                    Ok(()) => return Ok(()),
+                    Err(e) => {
+                        // If underlying layer doesn't support locking, fall through to fallback
+                        let errno: i32 = e.into();
+                        if errno != libc::ENOSYS {
+                            return Err(errno.into());
+                        }
+                    }
+                }
+            }
+        }
+
+        // Fallback: silently accept the lock request
+        Ok(())
     }
     /// check file access permissions. This will be called for the `access()` system call. If the
     /// `default_permissions` mount option is given, this method is not be called. This method is
--- a/project/libfuse-fs/src/passthrough/async_io.rs
+++ b/project/libfuse-fs/src/passthrough/async_io.rs
@@ -1875,31 +1875,88 @@ impl Filesystem for PassthroughFs {
     async fn getlk(
         &self,
         _req: Request,
-        _inode: Inode,
-        _fh: u64,
+        inode: Inode,
+        fh: u64,
         _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
+        start: u64,
+        end: u64,
+        r#type: u32,
+        pid: u32,
     ) -> Result<ReplyLock> {
-        Err(libc::ENOSYS.into())
+        if self.no_open.load(Ordering::Relaxed) {
+            return Err(enosys().into());
+        }
+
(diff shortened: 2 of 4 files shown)


project/libfuse-fs/src/passthrough/async_io.rs
```diff
@@ -1875,31 +1875,88 @@ impl Filesystem for PassthroughFs {
...
+        end: u64,
+        r#type: u32,
+        pid: u32,
     ) -> Result<ReplyLock> {
-        Err(libc::ENOSYS.into())
+        if self.no_open.load(Ordering::Relaxed) {
+            return Err(enosys().into());
+        }
+
+        let data = self.handle_map.get(fh, inode).await?;
```
genedna, line 1890: **`lock_owner` is unused.** POSIX `fcntl` `F_GETLK` operates per-process, not per-lock-owner, so two FUSE clients with different `lock_owner` values in the same fuse daemon process will see each other's locks. Consider using `F_OFD_GETLK` (Linux 3.15+) which operates per-fd and maps better to FUSE semantics. At minimum, add a comment explaining why `lock_owner` is intentionally ignored.

project/libfuse-fs/src/passthrough/async_io.rs
```diff
@@ -1875,31 +1875,88 @@ impl Filesystem for PassthroughFs {
...
     }
 
     #[allow(clippy::too_many_arguments)]
     async fn setlk(
         &self,
         _req: Request,
-        _inode: Inode,
-        _fh: u64,
+        inode: Inode,
+        fh: u64,
```
genedna, line 1926: Same concern as `getlk`: `_lock_owner` is unused, and `F_SETLK`/`F_SETLKW` are per-process locks. Using `F_OFD_SETLK`/`F_OFD_SETLKW` would provide per-fd lock semantics that align better with FUSE's lock_owner model.

project/libfuse-fs/src/overlayfs/async_io.rs
```diff
@@ -846,32 +846,102 @@ impl Filesystem for OverlayFs {
...
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
+        fh: u64,
+        lock_owner: u64,
```
genedna, line 852: **Mutex held across async call.** The `handles` `MutexGuard` is alive across the `.await` on `rh.layer.getlk(...)`. This blocks all concurrent handle operations. For `setlk` with `block=true` this could be unbounded. Consider cloning `rh.layer`, `rh.inode`, and `rh.handle` into locals, then dropping the guard before awaiting.

```rust
let (layer, real_inode, real_handle) = {
    let handles = self.handles.lock().await;
    // ... extract and clone ...
};
// guard dropped
layer.getlk(req, real_inode, real_handle, ...).await
```

project/libfuse-fs/src/overlayfs/async_io.rs
```diff
@@ -846,32 +846,102 @@ impl Filesystem for OverlayFs {
...
         &self,
-        _req: Request,
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
```
genedna, line 871: **Silent success may violate application expectations.** When `no_open` is true or there's no real handle, `setlk` returns `Ok(())` without actually acquiring any lock. Applications relying on file locks for mutual exclusion will silently break. Consider returning `Err(libc::ENOLCK.into())` or `Err(libc::ENOSYS.into())` when there's no real handle to delegate to.

project/libfuse-fs/src/unionfs/async_io.rs
```diff
@@ -841,32 +841,102 @@ impl Filesystem for OverlayFs {
...
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
+        fh: u64,
+        lock_owner: u64,
```
genedna, line 847: Same mutex-held-across-await concern as overlayfs. See comment on `overlayfs/async_io.rs`.

project/libfuse-fs/src/unionfs/async_io.rs
```diff
@@ -841,32 +841,102 @@ impl Filesystem for OverlayFs {
...
         &self,
-        _req: Request,
+        req: Request,
         _inode: Inode,
-        _fh: u64,
-        _lock_owner: u64,
-        _start: u64,
-        _end: u64,
-        _type: u32,
-        _pid: u32,
```
genedna, line 866: Same silent-success concern as overlayfs `setlk`. See comment on `overlayfs/async_io.rs`.

genedna (commented):
## Code Review: fix(libfuse-fs): improve async file-lock semantics and unify fallback behavior

Overall this is a solid improvement — replacing blanket `ENOSYS` stubs with real `fcntl`-based locking in passthrough and proper delegation in overlayfs/unionfs is the right direction. The ESTALE downgrade in `file_handle.rs` is also a sensible noise reduction. A few issues worth addressing before merge:

---

### 1. Security / Correctness: `lock_owner` is silently ignored (passthrough)

In `passthrough/async_io.rs`, the `_lock_owner` parameter is still prefixed with `_` and never used. POSIX `fcntl` `F_GETLK`/`F_SETLK` locks are per-process, not per-owner, so they don'\''t map cleanly to FUSE'\''s `lock_owner` semantics. This means:
- Two different FUSE clients (with different `lock_owner` values) sharing the same process will **interfere** with each other'\''s locks.
- `F_SETLK` unlock from one client may release a lock held by another.

This is a known limitation of POSIX locks, but it should at minimum be documented with a comment explaining the semantic gap. If correctness matters here, consider using `F_OFD_SETLK` / `F_OFD_GETLK` (open file description locks, Linux 3.15+) which are per-fd rather than per-process and map more naturally to FUSE lock_owner semantics.

### 2. Performance: Holding async mutex across I/O (overlayfs & unionfs)

In both `overlayfs/async_io.rs` and `unionfs/async_io.rs`, the `getlk` and `setlk` methods acquire `self.handles.lock().await` and hold the `MutexGuard` across the entire `.await` call to `rh.layer.getlk(...)` / `rh.layer.setlk(...)`. This blocks **all** other handle operations (open, read, write, release, etc.) for the duration of the underlying lock call — which, in the case of `setlk` with `block=true`, could be **indefinitely**.

**Recommendation:** Clone the needed data (`Arc<PassthroughFs>`, inode, handle) and drop the mutex guard before awaiting the underlying call.

### 3. Correctness: Silent success fallback in `setlk` may violate application expectations

When `no_open` is true, or when the handle/real_handle lookup fails, both overlayfs and unionfs `setlk` silently return `Ok(())`. This tells the application the lock was acquired, but **no lock was actually placed**. Applications relying on FUSE locking for mutual exclusion will see data corruption.

Consider returning an error (e.g., `ENOTSUP` or `ENOLCK`) when there is no real handle to delegate to, rather than pretending success. The ENOSYS fallback (underlying layer doesn'\''t support it) returning `Ok(())` is more defensible but should still be logged at `debug!` level.

### 4. Potential integer truncation in casts (passthrough)

The `as` casts in the `flock` construction could silently truncate:
- `r#type as libc::c_short` — if `type` has values beyond `i16` range
- `start as libc::off_t` — on 32-bit platforms, `u64` → `i64` can overflow
- `pid as libc::pid_t` — `u32` → `i32` wraps for large values

While unlikely in practice, using `try_into()` with proper error handling would be more robust.

### 5. Code duplication (overlayfs vs unionfs)

The `getlk` and `setlk` implementations in `overlayfs/async_io.rs` and `unionfs/async_io.rs` are **character-for-character identical**. Consider extracting the shared logic into a helper function or trait default method.

### 6. Testing

No tests are included for the new lock operations. Consider adding unit/integration tests for the fcntl path, the ENOSYS fallback, and `block=true` behavior.

### 7. ESTALE change (file_handle.rs) — Looks good ✓

Downgrading `ESTALE` from `error!` to `debug!` is appropriate.

---

| Area | Status |
|------|--------|
| Core approach | ✅ Good |
| ESTALE log level change | ✅ Good |
| ENOSYS fallback design | ⚠️ Silent success in setlk |
| lock_owner semantics | ⚠️ Needs documentation or OFD locks |
| Mutex held across await | ⚠️ Performance concern |
| Integer cast safety | ℹ️ Low risk, worth documenting |
| Code duplication | ℹ️ Consider refactoring |
| Test coverage | ❌ Missing |

jerry609:
lgtm

project/libfuse-fs/src/passthrough/async_io.rs
jerry609: make sense

project/libfuse-fs/src/passthrough/async_io.rs
jerry609: make sense

PR #515, opened 2026-03-09
fix(ci): add missing dependencies in release ci

erasernoob:


erasernoob:
This PR is moved to PR #516

PR #524, opened 2026-03-13
clean up fs.rs and unify etcd writes with EtcdTxn

mon3stera:
This PR refactors two related areas in SlayerFS: VFS handle access in fs.rs, and etcd write-side transaction
  handling.

  src/vfs/fs.rs had repeated handle lookup, stale-handle conversion, and direct metadata access patterns. This made
  the file noisy and increased the maintenance cost of otherwise simple changes. This PR extracts those repeated
  patterns into VFS/meta helpers so the call sites are shorter and more consistent.

  src/meta/stores/etcd_store.rs previously used multiple write-side patterns, including raw Txn::new(...), the old
  atomic_update helper, and ad hoc retry logic. That made etcd transaction behavior harder to reason about. This PR
  introduces EtcdTxn as the single transaction abstraction and migrates the remaining update paths to it.

  As a result, VFS code is cleaner, etcd writes now follow one consistent optimistic-transaction model, and future
  metadata changes should be easier to implement safely.

***This PR was fully handled by codex.***


genedna:
@claude

PR #540, opened 2026-03-17
Feature/slayerfs compact gc

Tyooughtul:
主要为slayerfs 添加gc和compact机制：
#### Compact 机制：
- 触发条件：可配置的阈值
- 两种模式：
  - Async Compact：slice count 在 [min_slice_count, sync_threshold) 范围内，后台执行，不阻塞写入
  - Sync Compact：slice count >= sync_threshold，先获取分布式锁，阻塞该 chunk 的写入，然后执行 compact

#### GC 机制
- 两阶段删除：
  - Phase 1: Soft Delete（标记 delayed，记录到 delayed_slice 表）
  - Phase 2: Hard Delete（GC worker 定期清理）

#### 存储层支持
- DatabaseStore：完整的 compact 和 GC 实现
- EtcdStore：支持分布式锁和 delayed slice 管理
- RedisStore：支持分布式锁和 delayed slice 管理

#### 测试覆盖
- `compaction_perf_test.rs`：性能测试
- `compactor_test.rs`：compactor 单元测试
- `gc_test.rs`：GC 单元测试
- `gc_compact_e2e_test.rs`：端到端集成测试
- `compaction_worker_test.rs`：background worker 测试

#355

### DIFF
--- a/project/slayerfs/doc/compact_gc.md
+++ b/project/slayerfs/doc/compact_gc.md
@@ -1,0 +1,390 @@
+# slice compact and gc design
+
+this document describes the slice compaction and garbage collection mechanisms in slayerfs.
+
+## overview
+
+slayerfs uses a write-once-read-many approach for data storage. when files are written, data is appended as slices within chunks. over time, this leads to:
+
+- multiple overlapping slices for the same data range
+- accumulation of outdated slice metadata
+- increased storage costs from orphaned objects
+
+the compact and gc mechanisms address these issues by merging overlapping slices and cleaning up obsolete data.
+
+## slice compact mechanism
+
+### purpose
+
+compact reduces metadata overhead and improves read performance by merging overlapping or adjacent slices within the same chunk.
+
+### when to compact
+
+compact is triggered based on configurable thresholds:
+
+| parameter          | default value | description                                    |
+| ------------------ | ------------- | ---------------------------------------------- |
+| min_slice_count    | 5             | minimum number of slices to trigger compact    |
+| min_fragment_ratio | 0.1           | minimum fragmentation ratio to trigger compact |
+| async_threshold    | 100           | slice count threshold for async compact        |
+| sync_threshold     | 350           | slice count threshold for sync compact         |
+
+compact is triggered when:
+
+- slice count >= min_slice_count (5) and
+- fragmentation ratio >= min_fragment_ratio (0.1)
+
+compact mode selection:
+
+- slice count >= min_slice_count (5) and < sync_threshold (350): async compact (background, non-blocking)
+- slice count >= sync_threshold (350): sync compact (intended to block writes to the chunk)
+
+note:
+
+- there are two modes (async/sync), not three. the async mode handles all cases from min_slice_count up to sync_threshold.
+- sync compact uses dynamic TTL-based locking (via `CompactLockManager`) to block writes during compaction
+
+### fragmentation ratio calculation
+
+```
+fragmentation_ratio = (total_slice_size - merged_slice_size) / total_slice_size
+```
+
+example:
+
+- slice a: offset 0, length 100
+- slice b: offset 50, length 100 (overlaps with slice a by 50 bytes)
+- total slice size: 200 (100 + 100)
+- merged slice size: 150 (0-150, after removing overlap)
+- fragmentation ratio: (200 - 150) / 200 = 0.25
+
+note: this measures the amount of overlapping data between slices, not the physical layout fragmentation. higher ratio means more redundant data that can be eliminated by compaction.
+
+### compact process
+
+1. identify chunks meeting compact thresholds
+2. select slices to compact => overlapping or adjacent
+3. merge slices into fewer, larger slices
+4. write new slices to object storage
+5. update metadata atomically => new slices in, old slices marked delayed
+6. old slices are soft-deleted => recorded in delayed_slice table
+
+### merge algorithm
+
+slices are merged by **removing only fully-covered slices**:
+
+- a slice is removed only when a **single newer slice** (higher `slice_id`) completely covers its `[offset, offset+length)` range
+- **partially overlapping slices are always kept intact** — their `offset` and `length` are never modified
+
+**why no splitting?** block data is stored at `(slice_id, block_index)` where `block_index` is relative to the slice's original `offset`. if a slice's `offset` or `length` were changed (e.g., by trimming the head), the block index mapping would break — reads would address the wrong blocks. example:
+
+```
+slice A: offset=0, length=200  →  blocks at (A,0), (A,1), ...
+if trimmed to offset=50, length=150  →  reader computes block_index relative to offset=50
+but the actual data is still stored at (A,0) which was relative to offset=0  →  wrong data!
+```
+
+this is why the algorithm only does **whole-slice removal**, never splitting/trimming.
+
+#### delayed data format
+
+when old slices are soft-deleted, they are serialized into a binary blob (20 bytes per slice):
+
+| field    | size    | encoding          |
+| -------- | ------- | ----------------- |
+| slice_id | 8 bytes | u64 little-endian |
+| offset   | 8 bytes | u64 little-endian |
+| size     | 4 bytes | u32 little-endian |
+
+this format is produced by `prepare_delayed_data()` and consumed by `cleanup_delayed_slices()`.
+
+## slice gc mechanism
+
+### purpose
+
+gc permanently removes obsolete slice metadata and cleans up corresponding object storage data after new slices are confirmed active.
+
+### two-phase deletion
+
+to ensure data safety, gc uses a two-phase deletion approach:
+
+#### phase 1: soft delete (delayed deletion)
+
+when compact replaces old slices:
+
+1. new slices are written and metadata committed
+2. old slices are recorded in `delayed_slice` table with timestamp
+3. old slices remain in `slice_meta` table (still readable)
+
+#### phase 2: hard delete (permanent deletion)
+
+gc worker periodically processes delayed slices:
+
+1. query delayed slices older than max_age_secs (default 1 hour)
+2. verify new slices are active and readable
+3. delete old slices from `slice_meta` table
+4. remove entries from `delayed_slice` table
+
+note: object storage data cleanup is handled separately by the object gc worker, which scans for orphaned objects not referenced by any slice metadata.
+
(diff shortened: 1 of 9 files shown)


project/rfuse3/src/raw/session/handlers.rs
mon3stera, line 1: Please open a separate PR for rfuse3 changes and explain what was changed and why.

.github/workflows/base.yml
mon3stera, line 1: Did you remove these configs by mistaken?

mon3stera (changes requested):

project/slayerfs/doc/compact_gc.md
```diff
@@ -0,0 +1,390 @@
...
+6. old slices are soft-deleted => recorded in delayed_slice table
+
+### merge algorithm
+
+slices are merged by **removing only fully-covered slices**:
+
+- a slice is removed only when a **single newer slice** (higher `slice_id`) completely covers its `[offset, offset+length)` range
+- **partially overlapping slices are always kept intact** — their `offset` and `length` are never modified
+
+**why no splitting?** block data is stored at `(slice_id, block_index)` where `block_index` is relative to the slice's original `offset`. if a slice's `offset` or `length` were changed (e.g., by trimming the head), the block index mapping would break — reads would address the wrong blocks. example:
```
mon3stera, line 79: The discussion about splitting is valid, provided that the data blocks aren't rewritten. However, I notice that the heavy compaction writes them. You need to mention it in this document.

project/slayerfs/src/chunk/compactor.rs
```diff
@@ -0,0 +1,402 @@
...
+                return Err(CompactorError::InvalidData(format!(
+                    "Slice {} exceeds chunk bounds: offset={}, length={}, chunk_size={}",
+                    slice.slice_id,
+                    slice.offset,
+                    slice.length,
+                    merged_data.len()
+                )));
+            }
+
+            merged_data[start..end].copy_from_slice(&data);
```
mon3stera, line 332: `block_store.read_range` directly reads data into the provided buffer. Therefore, you can eliminate an extra copy by splitting the buffer and passing them to `read_slice_data`.

project/slayerfs/doc/compact_gc.md
Tyooughtul: thanks for your comment, I will clearly distinguish between light and heavy paths in compact_gc.md to avoid misunderstandings later.

project/rfuse3/src/raw/session/handlers.rs
Tyooughtul: The session.rs consolidation was an intermediate step during slayerfs development for easier refactoring. Since those changes are now complete, I'm reverting rfuse3 back to its original structure to minimize unrelated modifications.

.github/workflows/base.yml
Tyooughtul: This is my ci testing process on the testing branch. To ensure that the ci can test my modifications normally, I made some changes, but I found that I seemed to have forgotten to change them back. Thank you for your reminder 😊. I'll re-check and push later.

project/slayerfs/src/meta/stores/etcd_store.rs
```diff
@@ -3415,6 +3446,360 @@ impl MetaStore for EtcdMetaStore {
...
+    ) -> Result<(), MetaError> {
+        if !old_slices_to_delay.is_empty() && !old_slices_to_delay.len().is_multiple_of(20) {
+            return Err(MetaError::Internal(
+                "Invalid delayed data length".to_string(),
+            ));
+        }
+
+        let slice_key = key_for_slice(chunk_id);
+
+        self.atomic_update(
```
mon3stera: Updates to slices are atomic, but the operation as a whole is not. If the process crashes after the slices have been successfully written, the delay slices will not be written properly, resulting in a leak.

project/slayerfs/src/meta/stores/etcd_store.rs
```diff
@@ -3415,6 +3446,360 @@ impl MetaStore for EtcdMetaStore {
...
+                "Invalid delayed data length".to_string(),
+            ));
+        }
+
+        let slice_key = key_for_slice(chunk_id);
+        let result = self
+            .atomic_update(
+                &slice_key,
+                |old: Vec<SliceDesc>| {
+                    if old.len() != expected_slice_count {
```
mon3stera: If `old.len() != expected_slice_count`, it only indicates that the current compacted data is stale. The problem is you just reject the commit and retry, without doing a new compaction. So you're just going around in circles.

project/slayerfs/src/meta/stores/database/mod.rs
```diff
@@ -2828,12 +3005,317 @@ impl MetaStore for DatabaseMetaStore {
...
+        txn.commit().await.map_err(MetaError::Database)?;
+        Ok(())
+    }
+
+    async fn replace_slices_for_compact_with_version(
+        &self,
+        chunk_id: u64,
+        new_slices: &[SliceDesc],
+        old_slices_to_delay: &[u8],
+        expected_slice_count: usize,
```
mon3stera, line 3078: Just using `expected_slice_count` as the condition isn't enough. You must pass the full original descriptions to do that.

project/slayerfs/src/meta/stores/etcd_store.rs
```diff
@@ -3415,6 +3446,360 @@ impl MetaStore for EtcdMetaStore {
...
+            }
+            let delayed_id = slice_id as i64; // Use slice_id as delayed_id for simplicity
+            let mut updated_value = value.clone();
+            if let Some(obj) = updated_value.as_object_mut() {
+                obj.insert("status".to_string(), serde_json::json!("meta_deleted"));
+                obj.insert("deleted_at".to_string(), serde_json::json!(Utc::now().timestamp()));
+            }
+            
+            client
+                .put(key.as_ref(), updated_value.to_string(), None)
```
mon3stera: This flow isn't transactionally safe. There are multiple crash and concurrency windows that can leave the system in partially applied states. Unless you can prove the protocol is correct under partial failure, retry, and duplicate execution, this implementation is strongly not acceptable.

mon3stera (changes requested):

project/slayerfs/src/meta/stores/etcd_store.rs
Tyooughtul: > If `old.len() != expected_slice_count`, it only indicates that the current compacted data is stale. The problem is you just reject the commit and retry, without doing a new compaction. So you're just going around in circles.

Let me clarify the retry mechanism: when ContinueRetry is returned, the compaction worker catches this error and continues to the next chunk. In the next cycle, it will re-fetch the latest slices via analyze_chunk and retry the compaction with fresh data. So it's not an infinite loop - just deferred retry to the next cycle.

project/slayerfs/src/meta/stores/etcd_store.rs
```diff
@@ -3415,6 +3446,360 @@ impl MetaStore for EtcdMetaStore {
...
+            return Err(MetaError::Internal(
+                "Invalid delayed data length".to_string(),
+            ));
+        }
+
+        let slice_key = key_for_slice(chunk_id);
+
+        self.atomic_update(
+            &slice_key,
+            |_old: Vec<SliceDesc>| Ok((new_slices.to_vec(), ())),
```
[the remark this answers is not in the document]
mon3stera: codex is right.

project/slayerfs/src/meta/stores/redis_store.rs
```diff
@@ -2068,6 +2076,508 @@ impl MetaStore for RedisMetaStore {
...
+                );
+            } else {
+                tracing::debug!(
+                    slice_id = slice_id,
+                    "Slice already deleted from chunk"
+                );
+            }
+            
+            let new_value = format!("{}:{}:{}:{}:meta_deleted", chunk_id, offset, size, created_at);
+            let _: () = redis::cmd("SET")
```
mon3stera: Same as etcd.

project/slayerfs/src/chuck/gc.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        let num_blocks = size.div_ceil(block_size);
+
+        if num_blocks == 0 {
+            return Ok(());
+        }
+        let start_block = offset / block_size;
+
+        self.block_store
+            .delete_range((slice_id, start_block as u32), num_blocks)
```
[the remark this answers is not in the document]
mon3stera: codex is right.

project/slayerfs/src/meta/stores/etcd_store.rs
mon3stera: Compaction worker cannot catch the `ContinueRetry`. `atomic_update` calls `backoff(f, max_retries)` at the end, so it is actually `backoff` that catches the error, but it doesn't pass `ContinueRetry`. Ultimately, the upper layer receives a `MaxRetriesExceeded`.

mon3stera (changes requested):

project/slayerfs/src/meta/stores/etcd_store.rs
Tyooughtul: You’re right about the current etcd behavior: stale checks (expected_count / expected slice set mismatch) currently go through the same internal backoff path, so they may retry up to 10 times and then surface as MaxRetriesExceeded.

I agree this is inefficient for stale-input conflicts. It is not a correctness bug or infinite loop, but retry waste.

I’ll adjust this path to fail fast on stale-compaction conflicts and reserve backoff retries for transaction/CAS conflicts only.

> Compaction worker cannot catch the `ContinueRetry`. `atomic_update` calls `backoff(f, max_retries)` at the end, so it is actually `backoff` that catches the error, but it doesn't pass `ContinueRetry`. Ultimately, the upper layer receives a `MaxRetriesExceeded`.

Let me clarify the semantics here.
In the etcd path, stale checks currently return through the internal retry path and may finally surface as MaxRetriesExceeded rather than raw ContinueRetry.

The key point is: this is not an infinite loop. Retries inside one call are bounded, and after failure the worker proceeds; later cycles re-read latest slices and run a fresh compaction attempt.

So there are two separate cases:
1、stale compaction input,
2、transient transaction/CAS conflicts.

Can I understand that your meaning is that if it's an expired issue, it should be recompact immediately instead of using the same meaning as 2?

PR #574, opened 2026-04-07
fix(slayerfs): remove duplicate etcd struct.

yyjeqhc:
Since the forward index supports prefix queries, we can list all child entries of a directory directly from `EtcdForwardEntry`. This makes `EtcdDirChildren` redundant, so this PR removes it.


Ivanbeethoven:
@codex review

Ivanbeethoven (approved):

PR #576, opened 2026-04-10
ci: migrate base workflow to the new runner scale set

jjl9807:


jjl9807:
The `rkforge CI` workflow was intentionally cancelled because it falls outside the scope of this PR's workflow migration.

genedna (approved):

PR #592, opened 2026-04-17
[slayerfs] fix(meta/cache): prevent creating incomplete cache from si…

Ivanbeethoven:
…ngle slices after truncate; fix xfstests read-bad-data

Update append_slice/replace_slices to only modify existing cache entries, avoiding insertion of incomplete per-chunk lists from a single slice.

Remove mount option custom_options(sync) (clean up an unnecessary workaround).

Fix cache race between truncate and subsequent writes that caused READ BAD DATA in xfstests generic/091 and generic/112; targeted tests have passed.


Ivanbeethoven:
@codex review

genedna (approved):

PR #611, opened 2026-05-01
feat(rfuse3) & feat(libfuse-rs)  ddress macFUSE startup review & add macOS support with conformance and benchmarks

genedna:


project/libfuse-fs/src/passthrough/config.rs
```diff
@@ -207,6 +242,11 @@ impl Default for Config {
             use_mmap: false,
```
genedna, line 241: **Config default vs. helper-default divergence is a footgun.**

`Config::default()` sets `macos_lazy_inode_fd: false`, but `new_passthroughfs_layer` overrides it to `true` (passthrough/mod.rs:107). The doc-comment on the field (config.rs:194-211) explicitly calls this out as intentional ("eager-mode tests stay opt-in"), but the support-matrix and roadmap claim "lazy is the default on macOS" — which is only true if you go through the helper.

Anyone who builds a custom `Config` directly (e.g. the new `--macos-eager` example path you added at examples/passthrough.rs:80, or any downstream caller wiring up `PassthroughFs::<()>::new(cfg)`) silently lands on the eager path with TTL=0 and very different performance characteristics. The phase-7 A/B numbers — 2.76× lazy/eager — would not surface for those callers.

Two reasonable options:
1. Flip `Config::default()` to `macos_lazy_inode_fd: true` and have eager-mode tests construct an explicit `Config { macos_lazy_inode_fd: false, ..Default::default() }`. Aligns the default with the documented platform behavior.
2. Keep the current divergence but rename the field to `macos_lazy_inode_fd_explicit: Option<bool>` so callers must consciously opt in or out, with `None` meaning "use the helper-resolved default." Avoids the silent eager fallback for direct `Config` users.

I lean toward (1) — the existing doc-comment is the only guard against this footgun, and the divergence already surprised me reading the diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/passthrough/mod.rs
```diff
@@ -204,34 +230,211 @@ impl AsFd for InodeFile<'_> {
...
+        }
+    }
+
+    /// Drop the entry for `inode` (called from `forget_one` once refcount
+    /// reaches zero). Idempotent — missing entries are a no-op. Does **not**
+    /// touch `ReopenableState::cached`; the inode itself is going away.
+    fn remove(&self, inode: Inode) {
+        let mut guard = self.inner.lock().expect("LazyFdLru mutex");
+        let _ = guard.pop(&inode);
+    }
```
genedna, line 348: **`std::sync::Mutex` poisoning could escalate one panic into a fs-wide outage.**

`LazyFdLru::touch`, `remove`, `len`, and every `ReopenableState` access uses `lock().expect("…")`. If any thread panics while holding one of these mutexes, every subsequent access poisons → panics → cascades up through the FUSE worker. On a long-running mount, a single bad inode can take down everyone reading any file.

In an async filesystem context, the safer pattern is:
```rust
let mut guard = self.inner.lock().unwrap_or_else(|e| e.into_inner());
```
The data is only an LRU map / a `(path, Option<Arc<File>>)` pair — there's nothing intrinsically inconsistent about it post-panic. Recovering avoids a fs-wide blast radius from a single failure.

(Same comment applies to all `expect("ReopenableState mutex")` and `expect("LazyFdLru mutex")` sites in this file — please consider sweeping them together.)

Bonus question: have you considered `parking_lot::Mutex` here? It's not poisonable, has a smaller footprint, and is what most async-heavy crates use. The "minimal dependency footprint" comment near `LazyFdLru` is a fine reason to stay on `std`, but worth weighing against the failure mode.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/passthrough/mod.rs
```diff
@@ -262,9 +465,28 @@ impl InodeHandle {
     fn do_stat(&self) -> Result<libc::stat> {
         match self {
             InodeHandle::File(f) => stat_fd(f, None),
-            InodeHandle::Handle(_h) => {
-                let file = self.get_file()?;
-                stat_fd(&file, None)
+            InodeHandle::Reopenable { state } => {
+                // Stat by path — no need to keep an fd around just for stat.
```
genedna, line 483: **`InodeHandle::get_file()` for `Reopenable` returns a runtime-error trapdoor.**

This is a programmer-error path: callers must use `InodeData::get_file` instead. Returning `io::Error::other("…")` means a bug only surfaces when someone exercises that exact code path (and probably only on macOS), and gets reported up the stack as a generic FUSE error to the kernel. The kernel will translate that into something like `EIO` for the userspace caller.

Two ways to make this catch earlier:
1. **`debug_assert!`/`panic!` in debug builds**: the comment says it's a programmer error, so make the failure mode loud during development. Production builds can keep the soft error to avoid taking the mount down.
2. **Type-system fix**: split `InodeHandle` into a public-facing version (no `Reopenable`) and an internal-only one. Or restructure so the only way to reach `Reopenable` is through `InodeData`, e.g. by making `InodeHandle::get_file` only callable from `InodeData::get_file` via a private trait.

Option (1) is the minimal change and probably enough; (2) is a deeper refactor and would be follow-up work.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/passthrough/async_io.rs
```diff
@@ -814,6 +831,70 @@ impl<S: BitmapSlice + Send + Sync> PassthroughFs<S> {
...
+                continue;
+            }
+            let Some(other_path) = other.lazy_path() else {
+                continue;
+            };
+            // `strip_prefix` requires a path-component match (won't match
+            // "/foo" against "/foobar"), which is exactly what we want.
+            if let Ok(suffix) = other_path.strip_prefix(&old_path) {
+                let mut rewritten = new_path.clone();
+                rewritten.push(suffix);
```
genedna, line 891: **`O(N_inodes)` directory-rename walk under read lock is a real risk under build-system workloads.**

The implementation is correct and the comment is honest about the trade-off, but worth pushing back on. Build systems (cargo, ninja, gradle) routinely do directory renames as part of atomic-publish patterns: build into `tmp.XXXX`, rename to `final-name`. With a moderately-sized workspace mounted, every such rename now walks every inode in the tree under a `RwLock` read.

Two concrete improvements that don't cost much:

1. **Skip the walk when the renamed entry is not a directory** — already done, good.
2. **Add a parent → children index in `InodeStore`** — `BTreeMap<Inode, Vec<Inode>>` updated in `do_lookup` (parent known) and cleared in `forget_one`. Then the rename walk becomes O(descendants), bounded by what the user actually has cached. For most workloads this is dramatically less than `N_inodes`.

The current code path also acknowledges this in the comment ("if it ever shows up in profiles, switch to an explicit parent→children index"). I'd push to do it now — once the lazy-fd path is the macOS default, you'll have to chase the regression report when it appears, which is harder than landing the index up front.

That said, this is not a blocker for the PR — the correctness story is solid. Filing a follow-up issue would be appropriate.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/passthrough/mod.rs
```diff
@@ -280,9 +502,54 @@ pub struct InodeData {
...
+/// Always sets `O_CLOEXEC`. Returns the raw fd; the caller wraps it in `File`.
+#[cfg(target_os = "macos")]
+fn lazy_open_path(path: &CStr, flags: libc::c_int) -> io::Result<libc::c_int> {
+    // Strip flags that don't make sense for this opener.
+    let base = (flags & !libc::O_CREAT & !libc::O_DIRECTORY) | libc::O_CLOEXEC;
+    let with_nofollow = base | libc::O_NOFOLLOW;
+    let fd = unsafe { libc::open(path.as_ptr(), with_nofollow) };
+    if fd >= 0 {
+        return Ok(fd);
+    }
```
genedna, line 561: **TOCTOU between the `O_NOFOLLOW` open and the `O_SYMLINK` retry.**

Between getting `ELOOP` on the first `open(O_NOFOLLOW)` and the retry with `O_SYMLINK`, the path can be replaced. If a regular file at `path` is replaced with a symlink between the two calls, the second open succeeds but on the wrong file. Conversely, a symlink replaced by a regular file would still work on the first open and never reach the retry.

Within the trust boundary of `cfg.root_dir` (which we canonicalize at startup, good), this is mostly an availability concern, not a security one — the lazy-fd path always opens paths under our own root. But it's worth noting in the doc comment, and worth thinking about whether the retry could be folded into a single `openat`-style call (it can't on macOS, hence why we're here at all).

A safer construction once you do canonicalize: stat the entry first (`fstatat(AT_SYMLINK_NOFOLLOW)`), and pick the open flags based on `S_ISLNK` rather than relying on a retry-on-error race. That removes the race window entirely at the cost of an extra syscall — but the extra syscall only happens on the first open per inode, since the result is cached in `ReopenableState`.

Not blocking; just a hardening note that would fit nicely with the future "canonicalize root_dir + use `O_NOFOLLOW_ANY`" work documented in the support matrix.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/unionfs/mod.rs
```diff
@@ -2166,6 +2230,89 @@ impl OverlayFs {
...
+            inode: entry.attr.ino,
+            whiteout: false,
+            opaque: false,
+            stat: Some(ReplyAttr {
+                ttl: entry.ttl,
+                attr: entry.attr,
+            }),
+        };
+        node.add_upper_inode(real, true).await;
+        Ok(Some(Arc::clone(node)))
```
genedna, line 2335: **`try_macos_apfs_clone_up` is duplicated wholesale between `overlayfs` and `unionfs`.**

The bodies of `overlayfs::OverlayFs::try_macos_apfs_clone_up` and this `unionfs::OverlayFs::try_macos_apfs_clone_up_unionfs` are 90% identical — same locking dance, same path joining, same fall-through error matrix. Duplication makes correctness drift hard to spot during follow-up changes; e.g. if the EEXIST-handling rule changes, both copies need to be edited.

The PR comment acknowledges this ("the unionfs layer trait carries `async_trait` bounds and a different `RealInode` shape, so they can't be folded into one helper without further refactoring"). That's fair, but the cost is small enough that it might be worth doing now: extract a free function `passthrough::util::try_apfs_clone_into(src_path, dst_dir, name) -> io::Result<bool>` and let each layer wire its own `RealInode` construction around it. The trait-bound difference is in the *call site*, not the clone helper itself.

If you'd rather not refactor in this PR (already at 4,700 LOC), please at least file a follow-up. The current state is fine to ship; the dup will just bite at the next maintenance touch.

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/overlayfs/layer.rs
```diff
@@ -26,106 +47,144 @@ pub trait Layer: Filesystem {
...
+                    let ie: std::io::Error = e.into();
+                    if ie.raw_os_error() == Some(libc::ENOENT) {
+                        Ok(false)
+                    } else {
+                        Err(ie.into())
+                    }
+                }
+            };
+        }
+
```
genedna, line 214: **Behavior divergence between `CharDev` and `OciWhiteout` in `delete_whiteout` for non-existent markers.**

The `CharDev` path looks up the entry first; if it returns `ENOENT`, the lookup error propagates back to the caller. Reading the original code, the `Err(e) => return Err(e)` arm in the CharDev path makes that a hard failure on a missing marker.

The `OciWhiteout` path goes straight to `unlink(parent, &.wh.name)` — which surfaces `ENOENT` from `unlink(2)` directly to the caller, but also any other unlink error. The caller can't distinguish "marker was already gone, harmless" from "marker exists but unlink failed" without inspecting the errno.

Two concrete things to consider:
1. **Decide what "delete a whiteout that doesn't exist" should do** and make both paths agree. Idempotent (`Ok(())` on `ENOENT`) is the friendlier semantic for overlayfs callers; if you go that way, both paths need to swallow `ENOENT`.
2. **Add a unit test** for the missing-marker case in both formats — currently the unit tests only cover present markers.

Same comment applies to `unionfs/layer.rs:140` (the parallel implementation).

---
_Generated by [Claude Code](https://claude.ai/code)_

project/libfuse-fs/src/passthrough/async_io.rs
```diff
@@ -2080,29 +2168,62 @@ impl Filesystem for PassthroughFs {
...
+            let st = stat_fd(&_fd, None)?;
+            let current_size = st.st_size as libc::off_t;
+            if target_size > current_size {
+                let mut store = libc::fstore_t {
+                    fst_flags: libc::F_ALLOCATEALL,
+                    fst_posmode: libc::F_PEOFPOSMODE,
+                    fst_offset: 0,
+                    fst_length: target_size - current_size,
+                    fst_bytesalloc: 0,
+                };
```
genedna, line 2212: **Race window between `stat_fd` and `F_PREALLOCATE`.**

Sequence:
1. `stat_fd(&_fd, None)?` → `current_size = X`
2. Another writer extends the file to size Y > X (no lock held).
3. We compute `fst_length = target_size - X`, but the actual hole-to-fill is `target_size - Y`, which could be negative.

`F_PREALLOCATE` with a zero or negative `fst_length` is benign on macOS — the kernel won't shrink, and an over-reserve still gets clamped by the next `ftruncate(target_size)`. So this is functionally correct, but the path *can* over-allocate on contended files. If `target_size <= current_size_at_call`, you're doing a no-op `F_PREALLOCATE` plus a possible no-op `ftruncate`.

Two minor cleanups:

1. Skip the whole block when `target_size <= current_size` — current code already does. Good.
2. Consider documenting that "any concurrent writer extending the file" leads to harmless over-reserve. Future readers will wonder.

Not a blocker. Real Linux `fallocate` has the same race-on-shared-fd hazard.

---
_Generated by [Claude Code](https://claude.ai/code)_

genedna (commented):
## Comprehensive review — macOS / macFUSE support for libfuse-fs + rfuse3

Reviewed all 37 files (4,746 additions / 485 deletions). Excellent work — this is the most thorough cross-platform port I've seen in this codebase: every macOS-specific kernel difference is documented, A/B-benchmarked, and given a kill-switch. Below is the structured feedback. **Overall posture: ready to merge once the sharp-edge items are addressed; the rest can be follow-ups.**

---

### 1. Code Quality

**What's working well:**
- The `unwrap()`-to-`ok_or_else(EINVAL)` sweep across `osstr_to_cstr` callers (passthrough/async_io.rs, overlayfs/mod.rs, unionfs/mod.rs) closes a real panic surface and is paired with a regression test (`lookup_rejects_nul_name_without_panicking`). Solid hygiene.
- `osstr_to_cstr_or_einval` helper keeps the conversion ergonomic. Good factoring.
- Removing the "fake success" behavior in `setxattr`/`removexattr` is a quiet but important correctness fix — `error!()`-then-`Ok(())` was actively misleading conformance suites.
- `WhiteoutFormat` abstraction (`util/whiteout.rs`) is clean, well-tested, and the per-platform `Default` impl makes the OCI-on-macOS choice explicit instead of buried in `cfg!` checks.
- Phase-by-phase tracking in `macos-roadmap.md` with explicit kill-switch criteria is exactly the kind of documentation that makes long-running cross-platform work auditable. The honest "actually 1.46× pre-A baseline, 2.76× same-environment A/B" debrief is much better than fabricating success.

**Sharp edges (see inline comments):**
- `Config::default()` vs. `new_passthroughfs_layer` divergence on `macos_lazy_inode_fd` — easy to land on eager mode by accident if you build a `Config` directly.
- `std::sync::Mutex` poisoning in `LazyFdLru` and `ReopenableState` — could turn one panic into a fs-wide outage.
- `InodeHandle::get_file()` returning a runtime error for `Reopenable` — programmer-error trapdoor; wants `debug_assert!`.
- `try_macos_apfs_clone_up` duplicated wholesale between `overlayfs` and `unionfs`. Refactor candidate.

**Minor nits not worth inline comments:**
- `passthrough/mod.rs:867` — `LazyFdLru::cap()` and `len()` use `#[allow(dead_code)]`. Consider gating to `#[cfg(test)]` or removing if there are public accessors (`macos_lazy_fd_cap`/`macos_lazy_fd_cache_len`) that already wrap them.
- `examples/passthrough.rs:79-115` — the `--macos-eager` branch reproduces a chunk of `new_passthroughfs_layer`'s timeout-zeroing. Could call into the helper with an `eager: bool` parameter to keep the example minimal.
- `tests/macos_mount_smoke.sh` — `mount | grep -q " $MOUNT_POINT_REAL "` is the standard-but-fragile mount detection idiom. On a path containing spaces this still works because of the surrounding spaces, but worth a comment for future maintainers.

### 2. Security

**Strong points:**
- `is_linux_only_xattr` correctly rejects `security.*`, `trusted.*`, `system.*` namespaces upfront — avoids leaking ENOTSUP info from the kernel and matches the Linux semantics the user expects.
- `validate_path_component` is called consistently before all rename/unlink/create paths.
- `cfg.root_dir` canonicalization at startup (when lazy mode is on) prevents stored paths from being symlink-resolved at use-time. Good defense-in-depth.
- `unsafe` blocks are limited to libc FFI, with adjacent error-return checks. No pointer arithmetic in user-supplied input territory.
- `try_apfs_clonefile` requires the destination not to exist (refuses to silently overwrite), and the test asserts EEXIST is surfaced as `Err`. Important — silent overwrite would be a data-loss vector.
- Test scripts use `safe_rm_dir` with explicit deny-list against `/`, `/tmp`, etc. Defensive cleanup is correct.

**Items to consider:**
- `lazy_open_path` two-step `O_NOFOLLOW` → `O_SYMLINK` retry has a TOCTOU window (inline comment). Confined to `cfg.root_dir`, so this is mostly availability, not classical TOCTOU privilege escalation. Worth noting in the doc-comment.
- `setxattr` for macOS now passes `_position` through to `fsetxattr` — only meaningful for `com.apple.ResourceFork`. The test (`macos_resource_fork_xattr_honors_position`) verifies the happy path, but doesn't verify that non-zero `_position` for *other* xattrs is rejected by the kernel (which would be the correct macOS behavior). Worth a one-line test if it's cheap.

### 3. Performance

**Solid evidence:**
- Same-environment A/B (phase 7): meta-stat 2.76× lazy/eager, randread-4k 1.42×, seqread-1m 1.26×. The honesty about earlier "concurrent_mounts" noise is great.
- `script/macfuse_bench.sh --ab` enforces `AB_META_MIN_RATIO=2.0` and `AB_IO_MAX_REGRESSION=0.10` as gates that fail the script. This is exactly the regression-blocking shape we want.
- `count_competing_macfuse_mounts` precondition prevents accidentally publishing noisy numbers.

**Concerns:**
- `macos_lazy_after_rename` performs an `O(N_inodes)` walk of the inode store on every directory rename (inline comment). This is the only scaling concern that bothers me — build systems do many directory renames, and at scale this read-locks the whole store.
- Global `Mutex<LruCache>` in `LazyFdLru::touch` is a single point of contention. Per-shard hashing or `dashmap`-based cache would scale better under high concurrency. Not blocking; observable in fio meta-stat at extreme thread counts.
- `copy_file_range` partial-copy fallback allocates a 64 KiB heap buffer per call. For workloads doing many small `copy_file_range`s, consider a per-task pool. Unlikely to be hot for typical container-image workloads.

### 4. Testing

**Strong:**
- `macos_lazy_dir_rename_rewrites_descendants` directly exercises the recursive-path-rewrite logic. Good targeted coverage.
- `macos_lazy_fd_pressure_caps_real_fds` is genuinely impressive — counts process fds via `/dev/fd` to assert the LRU caps real OS resource usage, not just an internal counter. This is the right way to test resource-bounding.
- `macos_apfs_clone_roundtrip` verifies both the byte-identical-output property and EEXIST behavior. Both are correctness-critical.
- `lookup_rejects_nul_name_without_panicking` is a regression guard for the `unwrap` → `EINVAL` sweep.
- The CI workflow has three jobs with the right risk gradient: `check` (must pass), `mount-smoke` (advisory, GH-hosted), `self-hosted-mount-smoke` (gated by repo var, required when present). Sensible separation.

**Gaps:**
- No stress test under low `RLIMIT_NOFILE` (e.g. 64). The fd-pressure test uses a synthetic cap but the real failure mode is a hostile rlimit.
- The `delete_whiteout` ENOENT semantics divergence (inline comment) lacks a test in either format.
- `force_readdir_plus` negotiation has unit tests (`force_readdirplus_does_not_advertise_unsupported_capability`) — good. No integration test that confirms a Linux client receives the expected reply flags. Probably out of scope for this PR.
- `macfuse_bench.sh --ab` has no test of its own. The script is large enough (~375 lines) that a unit test of the Python summary block would be reasonable. Out of scope here, but worth filing.

### 5. Documentation

**Excellent:**
- `macos-support-matrix.md` is a model of "what works, what doesn't, why" — every row links to either a code symbol or a tracking issue.
- `macos-conformance-baseline.md` documents the regression-blocking comparison with the exact pjdfstest pass/fail counts and history.
- `macos-performance-baseline.md` keeps a per-run diary that you can audit retroactively.
- Phase 6/7/8/9 sections in the support matrix make the rollout intent and decision history clear.

**Suggestions:**
- `macos-roadmap.md` is in Chinese. Given that this is a public open-source project under an English-speaking organization, this limits accessibility for non-Chinese-reading reviewers and contributors. Worth translating, even if it's auto-translated and human-verified.
- `Config::macos_lazy_inode_fd` doc-comment marks the field "experimental — semantics not API-stable" but the README/support-matrix says it's the macOS default. A quick "**Stability**: configuration knob is experimental; lazy-fd *behavior* is the production default" reconciliation would help.
- The `lru` workspace dependency is added in `Cargo.toml` for macOS only. Worth confirming `lru = "0.16"` (or whatever version) is defined in the workspace `Cargo.toml`'s `[workspace.dependencies]` — I couldn't see the workspace `Cargo.toml` in the diff. If you wired this up correctly elsewhere, ignore.

---

### Minor cosmetic items not worth inline comments

- `rfuse3/src/raw/session/handlers.rs:2483` — the patch removes a duplicate `#[cfg(target_os = "macos")]` attribute on `worker_setvolname`. Fixed in this PR. ✅
- `passthrough/mod.rs:457` — `do_stat` for `Reopenable` does an `fstatat(AT_FDCWD, abs_path, AT_SYMLINK_NOFOLLOW)`. Worth a comment that the `AT_FDCWD` is harmless because the path is always absolute.
- `examples/passthrough.rs:80` — `args.options.as_deref()` is parsed twice (once in the helper, once in the eager branch). Minor duplication.
- The TODO in `passthrough/async_io.rs:2455` (`// TODO: Switch to libc::renameat2`) is now obsolete on the Linux path — the rename2 method does dispatch to `renameat2`. Stale comment.

---

### Summary recommendation

Land the PR after addressing the **`Config::default()` divergence**, the **mutex poisoning** sweep, and the **`InodeHandle::get_file()` debug_assert** — those three are cheap and reduce real production-failure modes. Everything else can be a follow-up issue.

The phase-7 A/B numbers, the CI workflow design, and the kill-switch flag layout are all genuinely well-thought-out. Clear-eyed engineering — thanks for keeping the rollback story explicit.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #621, opened 2026-05-26
fix: nftables failing tests

Levi0804:
I was debug printing and testing some of the code to see how nftables creation and nftables in general are being handled in the repo. While testing, I noticed that these two tests were failing because they assumed that the chain name doesn't contain the protocol, but it actually does: 
https://github.com/rk8s-dev/rk8s/blob/aa0ccbc23b3c94688b06fcf781bd0f0efa18e507/project/libnetwork/src/nftables.rs#L803-L808

This PR fixes both of those failing tests.


genedna (approved):
Completed the nftables chain-name fix: the PR updated 2 tests; I updated the remaining services (svc-one/multi/reject/refresh/del/shrink/empty) to the proto-prefixed chain names that service_chain_name now produces. All 10 nftables_gen_check tests pass locally. Codex review: APPROVE. All code CI checks pass (Buck2 Build/Test, Clippy, Rustfmt, Redundancy). Merging via admin: 'Unit and Fast Integration Tests' is stuck on the unavailable self-hosted runner (infra), same as prior PRs.

Issue #615, opened 2026-05-11
rkl does not support port mapping feature


tcytan:
Description: 
Currently, the rkl component does not support port mapping functionality. Port mapping is important for real-world container use cases, such as exposing a port from the container to the host machine for external access.
Suggestions:  建议：

Add design and implementation for port mapping in rkl (e.g., syntax, configuration, parameters, etc.)
Support mapping single or multiple ports, including port ranges if possible
Update documentation to include usage examples and explanations

Levi0804:
Hi! I just wanted to make sure if my thinking is correct before I could proceed with a PR. 

For port mapping we could start by exposing this functionality through `rkl` cli (for both container workloads), letting users to configure how they want to map the ports. These ports could then be propagated to `rks`. Our main node (`rks`) will hold all the port mappings (probably inside the `Config` struct) and will then port forward (`NAT` it) the incoming requests appropriately to reach the desired location (container). Does all of this sound about right?

Also, could you please point out the location where we're actually handling the incoming requests on our main Node (`rks`)? thanks!

tcytan:
Maybe you can refer to the following content
rkl already parses hostPort / containerPort from YAML into PodSandboxConfig.port_mappings (task.rs:189-206), but never consumes them — no iptables/nftables DNAT rules are created. The CNI plugin chain only handles veth + bridge + IPAM.

  Changes

  1. New module — project/rkl/src/network/port_forward.rs

  pub fn apply_port_mappings(port_mappings: &[PortMapping], container_ip: &str) -> Result<()>
  pub fn remove_port_mappings(port_mappings: &[PortMapping], container_ip: &str) -> Result<()>

  Reuse libnetwork's existing nftables infrastructure (libnetwork/src/nftables.rs). Add a host_ports verdict map in the rk8s table. For each mapping, create a DNAT chain: host_ip:host_port → container_ip:container_port.

  2. Hook into pod network setup — task.rs

  After setup_pod_network() returns the pod IP (~line 366), call:
  port_forward::apply_port_mappings(&self.sandbox_config.port_mappings, &pod_ip)?;

  Call remove_port_mappings() on pod teardown.

  3. Hook into container path — commands/container/mod.rs

  After retrieve_container_ip() in start_container(), extract ports from the container spec and apply mappings. Clean up in delete_container().

  4. Port conflict detection

  Check for host_ip:host_port collisions in apply_port_mappings(), return a clear error on conflict.

Levi0804:
[...]

PR #619, opened 2026-05-20
clear up rkl.

Cosh-y:
Remove the logic in rkl that manages resources on a per-container basis and the logic that directly modifies cluster state without going through rks.

PR #622, opened 2026-05-29
feature(rkl): add port mapping

Levi0804:
This PR is WIP and aims to add port mapping feature to rkl (#615)


Levi0804:
@tcytan please let me know if these changes look good to you. Also, please let me know if I should squash my commits.

genedna:
## Conflict resolution + partial review fixes (maintainer)

I rebased this PR onto current `main` (it collided with #619, which removed the standalone `ContainerRunner` runtime that the original port-mapping hooks in `commands/container/mod.rs` were built on). The feature is now re-homed on the surviving `TaskRunner`/pod path. A Codex review then surfaced several correctness issues; I fixed the clear-cut ones in `38dfd888`:

- **`create_pod_sandbox_config`**: only declared host ports (`hostPort > 0`) become host mappings — a default `hostPort: 0` no longer creates an invalid nft host mapping. *(was a blocker)*
- **Static pods**: `sync_run_pod_sandbox` now also calls `apply_port_mappings` (previously only the async `run_pod_sandbox` was wired, so static pods got no mappings).
- **`stop_pod_sandbox`**: `remove_port_mappings` now runs even if `kill()` fails (was an early `?` return → leaked mappings on dead sandboxes).
- **Collision check**: keyed on `(protocol, host_ip, host_port)` so the same host ip/port on TCP+UDP isn't falsely rejected.

### ⚠️ Remaining issues to resolve before this can merge (deferred to you)

These need a design pass on the dnat chain lifecycle and the service-sync interaction:

1. **dnat chain naming is a process-local counter** (`port_forward.rs` `create_unique_chain_name` / `COUNTER`). `remove_port_mappings` deletes `dnat_0..COUNTER`, which (a) removes **other pods'** chains, and (b) cannot clean up after a daemon restart (counter resets). Use a **deterministic per-mapping chain name** (e.g. derived from proto+host_ip+host_port) and delete exactly those.
2. **Base `dnat` chain is `Add`ed on every apply** — the 2nd pod with host mappings fails on the already-existing chain/rule. Make base-chain/verdict-map creation idempotent (create-if-absent).
3. **Delete ordering**: `remove_port_mappings` deletes chains before the `host_ports` map elements that reference them — delete the map elements first.
4. **Teardown coverage**: normal pod deletion goes through `standalone.rs` (daemon delete + static-pod removal) and **bypasses `stop_pod_sandbox`**, so cleanup never runs on the common path. Wire `remove_port_mappings` into that delete path (best done together with the deterministic-naming change in #1 so it can target the right chains).
5. **Service full-sync wipes port mappings**: `libnetwork` full-sync flushes the shared `rk8s` table, and `nftrules_controller` doesn't replay the local `host_ports` entries / `dnat` chain. Port mappings get erased on the next service sync. Needs the port-forward state to be re-applied (or moved out of the flushed table).
6. *(minor)* `apply_port_mappings` failure leaves the pause container + pod network up with no rollback.

Conflict resolution + items above verified locally with `cargo check --tests`, `cargo fmt --check`, `cargo clippy -p rkl -p libnetwork -- -D warnings`. Holding the merge until 1–5 are addressed.

genedna (approved):

PR #626, opened 2026-07-13
rfuse3: account for padding when checking directory entries

vgough:
Correct readdir/readdirplus bounds checks to include alignment padding and avoid overflow, with tests for page boundaries.

### DIFF
--- a/project/rfuse3/src/raw/session/handlers.rs
+++ b/project/rfuse3/src/raw/session/handlers.rs
@@ -24,6 +24,7 @@ use crate::raw::reply::ReplyXAttr;
 use crate::raw::request::Request;
 use crate::SetAttr;
 
+use super::directory_entry_fits;
 use super::utils::{apply_direct_io, reply_error_in_worker, spawn};
 use super::worker::{DispatchCtx, WorkItem};
 pub(super) async fn worker_lookup<FS: Filesystem + Send + Sync + 'static>(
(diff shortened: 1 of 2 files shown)


Ivanbeethoven:
Thanks for the fix and the detailed regression coverage.

We recently extracted the `rfuse3` implementation from the rk8s repository into the standalone [`brewfs/async-fuse`](https://github.com/brewfs/async-fuse) project. BrewFS and other consumers now use it through the published `asyncfuse` crate.

The core directory-entry padding issue addressed by this PR is already functionally fixed in `asyncfuse 0.0.10`: the regular and worker `readdir` and `readdirplus` paths include alignment padding when checking the kernel-requested buffer size.

However, this PR still contains useful improvements that are not present in `0.0.10`, particularly the checked arithmetic, shared bounds helper, and page-boundary regression tests. We would like to move those improvements into the standalone async-fuse repository rather than maintain another copy in the old vendored implementation. If possible, please retarget or submit the relevant changes to [`brewfs/async-fuse`](https://github.com/brewfs/async-fuse). We can also port them while preserving attribution.

The inode-management issues described in [`rk8s-dev/rk8s#627`](https://github.com/rk8s-dev/rk8s/pull/627) are also valid and are not yet fixed in the standalone library. In particular, `path::InodePathBridge` still needs correct lookup-reference tracking, non-recycled node IDs, atomic bidirectional mappings, and stable rename/hard-link behavior. Those changes should likewise be merged into the standalone library, followed by an `asyncfuse 0.0.11` release and an rk8s dependency update.

Therefore, this PR should be considered superseded by the async-fuse extraction rather than rejected as invalid. Once the remaining hardening from this PR and the inode fixes from #627 have been moved to `brewfs/async-fuse`, the rk8s PRs can be closed in favor of the standalone implementation.

vgough:
> We recently extracted the rfuse3 implementation from the rk8s repository into the standalone [brewfs/async-fuse](https://github.com/brewfs/async-fuse) project. BrewFS and other consumers now use it through the published asyncfuse crate.

Great, thanks for letting me know. This will be easier to work on than as part of the larger rk8s. 
I've created a PR for the second PR (inode management), and will follow up after I've switched my code branch to using async-fuse and run more tests.

Feel free to close these PRs.

vgough:
After checking, the directory padding issue has not been fixed in asyncfuse. Without the directory entry fix, it is still not possible to build rust projects within a virtual filesystem. I'll submit a new PR to asyncfuse.

PR #630, opened 2026-09-23
libfuse-fs: unionfs no_open write fallback fix, Antares layer API, 0.3.0

Ivanbeethoven:
## Summary

Fixes unionfs write handling for **no_open-style FUSE requests**, where the kernel skips `OPEN` (via `FUSE_NO_OPEN_SUPPORT`) and issues `write` with `fh=0`. This broke `O_APPEND` writes to not-yet-copied-up lower files with `ENOENT`, which blocked ScorpioFS's MST/2 lower integration (gitmono-dev/scorpiofs#65). `libfuse-fs 0.3.0` was published to crates.io from this branch.

## The bug

With `FUSE_NO_OPEN_SUPPORT` negotiated, the kernel does not send `OPEN` for some write patterns (notably `O_APPEND` / partial-mode writes). The resulting `write` carries `fh=0`:

1. unionfs `get_data` falls back to the no_open path: it reconstructs a `HandleData` by looking up the node, copying it up if needed, and resolving the first-layer inode.
2. The fallback constructed `RealHandle.handle = 0` — a placeholder fh, not a real open.
3. The upper passthrough layer locates its file by **its own fh table**: `pread/pwrite(raw_fd)` keyed by fh → `0` resolves to nothing → `EBADF`, surfaced to the user as `ENOENT`.

So the very first append to a lower file failed. Once the file had been opened normally (e.g. by `truncate` or a read that triggered copy-up with a real OPEN), writes worked — which made the failure intermittent and hard to bisect.

## The fix

**`unionfs` fallback now opens the real layer file.** When `get_data` misses the fh, the fallback path resolves the target layer inode (`copy_node_up` when a write implies copy-up) and calls `layer.open` to obtain a genuine fh, then builds `HandleData` around it.

**Ephemeral handle lifecycle.** Because the kernel never sent `OPEN`, there will be no matching `RELEASE`, and registering these handles in the shared handle table caused slot collisions when concurrent no_open I/O streams reused slot 0 (earlier handles evicted mid-flight). Instead:

- fallback handles carry `ephemeral: true` and are **not** registered in the handle table
- `read`/`write` release the layer fd after the I/O completes (both in `mod.rs` and `async_io.rs`)

**Write body correction** (`async_io.rs`): the write body now honors the resolved `real_handle` (returns `ENOENT` only when no layer handle could be established, with an error log), instead of silently assuming slot 0.

## Verification

- ScorpioFS with MST/2 lower (crates.io `libfuse-fs 0.3.0`): `O_APPEND` write to a not-copied-up lower file → previously `ENOENT`, now succeeds; full append → sync → new snapshot round-trip green across 3 consecutive stability runs, including chunked reads of multi-hundred-KB blobs.
- Existing unionfs behavior unchanged for normal OPEN'd writes.

## Notes for reviewers

- `get_data` fh-miss handling was unified: the old `no_open ? fallback : fh-table lookup` split is gone; any fh miss goes through the same reconstruct path, which keeps `FLUSH`/`RELEASE` consistent since handles used for I/O are now real opens.
- The fallback is write-path-relevant only when the node's first layer is the upper rw layer; pure-lower reads keep the previous cheap path.

Refs: gitmono-dev/scorpiofs#65 (merged), crates.io `libfuse-fs 0.3.0`


Ivanbeethoven:
Closing this upstream PR. The work continues in the standalone repo gitmono-dev/libfuse-rs (full libfuse-fs history, including all the unionfs no_open fixes) and on crates.io (libfuse-fs 0.3.0). Thanks for the reviews.

